PR #1, opened 2025-03-27
feat: git operations for neuro

KTrain5169:
# ⚠️ This PR is a WIP ⚠️

Neuro can now use the Git extension within Visual Studio Code to perform Git operations, including committing, creating branches, interacting with remote repos, and more.

## Permissions

NeuroPilot will __not__ allow Neuro to perform git operations until the `Git Operations` permission is enabled.
Additionally, to allow her to perform operations on git configs and remotes, the `Git Configs` and `Git Remotes` permissions must also be enabled. All new permissions will require reloading permissions before changes take effect.

## New actions
During the period that the PR is a draft, this list may not be up-to-date. You'll probably have more of an up-to-date list by looking at the diffs.

### Normal git operations

- Adding files to index
- Reverting files
- Make commits
- Create, switch, rename and delete branches
- Switch between branches
- Review status of local git repo

### Configuration of git
Note: This is only for the local repository. Neuro cannot edit global git configurations, to prevent her from making Vedal accidentally impersonate someone.

- Initialise a git repo
- Get a config or all configs
- Set a config value

### Interact with remote repositories

- Add, get, change, and remove remotes
- Fetch, pull and push commits

## Other changes

- Removed .gitignore in playground(?)
- Split action functions in unsupervised.ts into their own files.


src/unsupervised.ts
```diff
@@ -185,751 +449,20 @@ export function registerUnsupervisedActions() {
...
-        if(!vscode.workspace.getConfiguration('neuropilot').get('permission.runTasks', false)) {
+        if (actionName === 'request_cookie') {
             return;
-        }
-
-        NEURO.client?.registerActions(NEURO.tasks.map((task) => {
-            return {
-                name: task.id,
-                description: task.description,
+        } else if (NEURO.tasks.find(task => task.id === actionName)) {
```
Pasu4, line 457: This condition is the wrong way around.

src/unsupervised.ts
```diff
@@ -185,751 +449,20 @@ export function registerUnsupervisedActions() {
...
-                    task: task,
-                });
-            }
-            else {
-                logOutput('INFO', `Ignoring task: ${task.name}`);
-            }
-        }
-
-        if(!vscode.workspace.getConfiguration('neuropilot').get('permission.runTasks', false)) {
+        if (actionName === 'request_cookie') {
```
Pasu4, line 455: We would need to exclude all other actions (from chat.ts and completions.ts) here as well. I think it would be better to handle this a different way:
1. Register a first handler in `NeuroClient.onAction` that sets a boolean `NEURO.actionHandled` to false.
2. In this handler and the other handlers from chat.ts / completions.ts / etc., set `NEURO.actionHandled` to true when the action is handled.
3. Register a last handler that checks `NEURO.actionHandled` and if false, sends a failed action result.

src/unsupervised.ts
```diff
@@ -23,9 +81,210 @@ export function registerUnsupervisedActions() {
...
+                    name: 'remove_git_remote',
+                    description: 'Remove a remote from the Git repository',
+                    schema: {
+                        type: 'object',
+                        properties: {
+                            remoteName: { type: 'string' },
+                        },
+                        required: ['remoteName'],
+                    }
+                },
```
Pasu4, line 260: I'm not sure if this is a good idea &mdash; Neuro would probably delete the remote and then forget the URL, and Vedal would have to fix it every time. I'd say we either move these two actions to a new permission or remove them altogether. There is probably not much she could do with it anyway, since she would likely have only a single remote she could even push to.

src/unsupervised.ts
```diff
@@ -23,9 +81,210 @@ export function registerUnsupervisedActions() {
...
+                    required: ['filePath']
+                }
+            },
+            {
+                name: 'delete_git_branch',
+                description: 'Delete a branch in the current Git repository',
+                schema: {
+                    type: 'object',
+                    properties: {
+                        branchName: { type: 'string' },
```
Pasu4, line 161: If possible/necessary with the git API, this should have an optional `force` property, since git doesn't let you delete a branch if it is not merged.

src/unsupervised.ts
```diff
@@ -23,9 +81,210 @@ export function registerUnsupervisedActions() {
...
+                    name: 'push_git_commits',
+                    description: 'Push commits to the remote repository',
+                    schema: {
+                        type: 'object',
+                        properties: {
+                            remoteName: { type: 'string' },
+                            branchName: { type: 'string' },
+                            forcePush: { type: 'boolean' },
+                        },
+                        required: ['remoteName', 'branchName']
```
Pasu4, line 281: `remoteName` and `branchName` should default to the default remote and current branch and not be required (like you can run `git push` without other arguments).

Pasu4 (changes requested):

src/unsupervised.ts
```diff
@@ -1,12 +1,70 @@
...
+import { handleTerminateTask, handleRunTask } from './tasks';
+import { handleCreateFile, handleCreateFolder, handleDeleteFileOrFolder, handleRenameFileOrFolder, handleGetFiles, handleOpenFile } from './file_actions';
+import { handleGetCursor, handleDeleteText, handleInsertText, handlePlaceCursor, handlePlaceCursorAtText, handleReplaceText } from './editing';
 
 /**
  * Register unsupervised actions with the Neuro API.
  * Will only register actions that the user has given permission to use.
  */
+
+const gitFiles: { [key: string]: (actionData: any) => void } = {
```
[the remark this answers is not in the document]
KTrain5169: I left it as `gitFiles` because my brain left at that moment, thx for pointing that out

src/git.ts
```diff
@@ -0,0 +1,485 @@
...
+ * Actions with the Git repo
+ * Requires neuropilot.permission.gitConfig to be enabled.
+ */
+
+export function handleNewGitRepo(actionData: any) {
+    if (!git) {
+        NEURO.client?.sendActionResult(actionData.id, false, 'Git extension not available.');
+        return;
+    }
+
```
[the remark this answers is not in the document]
KTrain5169: Right, I'll add that to every command.

src/unsupervised.ts
KTrain5169: I'll probs make them a new permission

KTrain5169: Implemented them locally, about to commit and push

src/unsupervised.ts
KTrain5169: Resolved locally

src/unsupervised.ts
Pasu4: Implemented in dev branch

src/git.ts
```diff
@@ -0,0 +1,612 @@
...
+                } else {
+                    NEURO.client?.sendActionResult(actionData.id, false, 'Both ref1 and ref2 are required for diffBetween.');
+                }
+                break;
+            
+            case 'fullDiff':
+                repo.diff().then((diff: string) => {
+                    NEURO.client?.sendActionResult(actionData.id, true, `Full diff:\n${diff}`);
+                });
+                break;
```
Pasu4, line 435: Action results should not be sent asynchronously, send a successful result after validating required parameters then send the result of the asynchrounous operation as context.

src/git.ts
```diff
@@ -0,0 +1,612 @@
...
+        return;
+    }
+    if (!vscode.workspace.getConfiguration('neuropilot').get('permission.gitOperations', false) && (!vscode.workspace.getConfiguration('neuropilot').get('permission.gitRemotes', false) && (!vscode.workspace.getConfiguration('neuropilot').get('permission.editRemoteData', false)))) {
+        NEURO.client?.sendActionResult(actionData.id, true, 'You do not have permission to perform Git operations.');
+        return;
+    }
+
+    const repo = git.repositories[0];
+    const remoteName: string = actionData.params?.remoteName;
+    const remoteUrl: string = actionData.params?.remoteUrl;
```
Pasu4, line 541: Property name in the schema is `remoteURL`.
```suggestion
    const remoteUrl: string = actionData.params?.remoteURL;
```

src/git.ts
```diff
@@ -0,0 +1,612 @@
...
+
+    if (!remoteName || !remoteUrl) {
+        NEURO.client?.sendActionResult(actionData.id, false, 'Remote name or URL missing.');
+        return;
+    }
```
Pasu4, line 546: Check separately to improve feedback.
```suggestion
    if (!remoteName) {
        NEURO.client?.sendActionResult(actionData.id, false, 'Missing required parameter "remoteName"');
        return;
    }
    if (!remoteUrl) {
        NEURO.client?.sendActionResult(actionData.id, false, 'Missing required parameter "remoteURL"');
        return;
    }
```

src/unsupervised.ts
```diff
@@ -1,34 +1,275 @@
...
+                                newRemoteName: { type: 'string' },
+                            },
+                            required: ['oldRemoteName', 'newRemoteName'],
+                        }
+                    },
+                    {
+                        name: 'get_git_remotes',
+                        description: 'Get a list of remotes in the Git repository',
+                        schema: {}
+                    }
```
Pasu4, line 249: Action has no associated handler.

src/git.ts
```diff
@@ -0,0 +1,612 @@
...
+
+    const repo = git.repositories[0];
+    const remoteName: string = actionData.params?.remoteName;
+    const branchName: string = actionData.params?.branchName;
+    const forcePush: boolean = actionData.params?.forcePush || false;
+
+    NEURO.client?.sendActionResult(actionData.id, true)
+
+    repo.push(remoteName, branchName, forcePush).then(() => {
+        NEURO.client?.sendContext(`Pushed commits${remoteName && ` to remote "${remoteName}"`}${branchName && `, branch "${branchName}"`}.${forcePush === true && " (forced push)"}`);
```
Pasu4, line 517: Results in "Pushed commitsundefinedundefined.false" if no parameters are set.

src/unsupervised.ts
```diff
@@ -190,753 +436,14 @@ export function registerUnsupervisedActions() {
...
-    edit.delete(document.uri, new vscode.Range(document.positionAt(textStart), document.positionAt(textStart + textToDelete.length)));
-    vscode.workspace.applyEdit(edit).then(success => {
-        if(success) {
-            logOutput('INFO', `Deleting text from document`);
-            vscode.window.activeTextEditor!.selection = new vscode.Selection(document.positionAt(textStart), document.positionAt(textStart));
-        }
-        else {
-            logOutput('ERROR', 'Failed to apply text deletion edit');
-            NEURO.client?.sendContext('Failed to delete text');
+        if(NEURO.tasks.find(task => task.id === actionData.name)) {
```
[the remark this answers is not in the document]
KTrain5169: I think implementing both a warning and the else should be what to do

src/git.ts
KTrain5169: I thought I changed this last commit? Might have been an issue with git merging, which could also explain like most other problems here

src/git.ts
```diff
@@ -0,0 +1,612 @@
...
+    }
+    if (!vscode.workspace.getConfiguration('neuropilot').get('permission.gitOperations', false)) {
+        NEURO.client?.sendActionResult(actionData.id, true, 'You do not have permission to perform Git operations.');
+        return;
+    }
+
+    const repo = git.repositories[0];
+    
+    NEURO.client?.sendActionResult(actionData.id, true)
+    repo.status().then((status: any) => {
```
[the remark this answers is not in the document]
KTrain5169: Wonder how we would incorporate that then. I sswear it worked as intended as well?

src/unsupervised.ts
```diff
@@ -1,34 +1,275 @@
...
+                name: 'git_remove',
+                description: 'Remove a file from the Git index',
+                schema: {
+                    type: 'object',
+                    properties: {
+                        filePath: { type: 'string' },
+                    },
+                    required: ['filePath']
+                }
+            },
```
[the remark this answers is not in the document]
KTrain5169: Shouldn't have been there in the first place, mb

src/git.ts
```diff
@@ -1,43 +1,292 @@
...
+                                    oldRemoteName: { type: 'string' },
+                                    newRemoteName: { type: 'string' },
+                                },
+                                required: ['oldRemoteName', 'newRemoteName'],
+                            }
+                        }
+                    ]);
+                }
+            }
+        }
```
Pasu4, line 289: Closing bracket for the suggested change above.
```suggestion
        });
```

src/git.ts
```diff
@@ -353,97 +760,241 @@ export function handleGitDiff(actionData: any) {
...
     const diffType: string = actionData.params?.diffType || 'diffWithHEAD'; // Default to diffWithHEAD
 
+    if (!['diffWithHEAD', 'diffWith', 'diffIndexWithHEAD', 'diffIndexWith', 'diffBetween', 'fullDiff'].includes(diffType)) {
+        NEURO.client?.sendActionResult(actionData.id, false, 'Invalid diffType.');
+        return;
+    } else {
+        NEURO.client?.sendActionResult(actionData.id, true);
+    }
+
     try {
```
Pasu4, line 781: Promise rejections are not caught by this try-catch and cause the extension to crash. We need an error handler in every `then` call here instead.

src/git.ts
```diff
@@ -353,97 +760,241 @@ export function handleGitDiff(actionData: any) {
...
+
+    repo.log().then((commits: Commit[]) => {
+        NEURO.client?.sendContext(`Commit log: ${JSON.stringify(commits)}`)
+    }, (err: string) => {
+        NEURO.client?.sendContext("Failed to get git log.")
+        logOutput("ERROR", `Failed to get git log: ${err}`)
+    })
+}
+
+export function handleGitBlame(actionData: any) {
```
Pasu4, line 888: No action is registered with the API for this handler, the handler is not registered in `gitActionHandlers` and the action is not documented in README.md.

Pasu4 (changes requested):

src/git.ts
```diff
@@ -1,43 +1,292 @@
...
+    'diff_files': handleGitDiff,
+    'merge_to_another_branch': handleGitMerge,
+    'abort_merge': handleAbortMerge,
+    'git_log': handleGitLog,
+    'tag_head': handleTagHEAD,
+    'delete_tag': handleDeleteTag
+};
+
+// Get the current Git repository
+let repo = git.repositories[0];
```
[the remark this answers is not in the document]
KTrain5169: The if conditional for (!repo) at the start of the function would've already caught this, no?

src/git.ts
```diff
@@ -1,43 +1,292 @@
...
+    if(vscode.workspace.getConfiguration('neuropilot').get('permission.gitOperations', false)) {
+        NEURO.client?.registerActions([
+            {
+                name: 'init_git_repo',
+                description: 'Initialize a new Git repository in the current workspace folder',
+                schema: {}
+            }
+        ]);
+
+        if (repo) {
```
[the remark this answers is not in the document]
KTrain5169: Works on my machine though...

src/git.ts
KTrain5169: I am actually forgetful, omg how did I not catch this

src/git.ts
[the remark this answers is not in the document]
Pasu4: It's better to do it this way because then the type checker will be able to highlight where we forget to check it.

src/git.ts
```diff
@@ -0,0 +1,1264 @@
...
+        if (invalidCommitOptionCheck === true) {
+            NEURO.client?.sendActionResult(actionData.id, false, `Invalid commit options: ${invalidCommitOptions.map((option: string) => {option})}`)
+            return;
+        }
+    }
+
+    NEURO.client?.sendActionResult(actionData.id, true)
+
+    repo.inputBox.value = message;
+    repo.commit(message).then(() => {
```
Pasu4, line 705: `ExtraCommitOptions` isn't being used.
```suggestion
    repo.commit(message, ExtraCommitOptions).then(() => {
```

src/git.ts
```diff
@@ -0,0 +1,1264 @@
...
+
+    NEURO.client?.sendActionResult(actionData.id, true)
+
+    repo.merge(refToMerge).then(() => {
+        NEURO.client?.sendContext(`Cleanly merged ${refToMerge} into the current branch.`)
+    }, (err: string) => {
+        NEURO.client?.registerActions([
+            {
+                name: 'abort_merge',
+                description: 'Aborts the current merge oepration.',
```
Pasu4, line 742: Typo
```suggestion
                description: 'Aborts the current merge operation.',
```

src/git.ts
```diff
@@ -0,0 +1,1264 @@
...
+    }
+    if (!repo) {
+        NEURO.client?.sendActionResult(actionData.id, true, 'You are not in a git repository.');
+        return;
+    }
+
+    NEURO.client?.sendActionResult(actionData.id, true)
+
+    repo.mergeAbort().then(() => {
+        NEURO.client?.unregisterActions(["abort_git_merge"])
```
Pasu4, line 768: Incorrect action name
```suggestion
        NEURO.client?.unregisterActions(["abort_merge"])
```

Pasu4 (changes requested):

src/git.ts
```diff
@@ -0,0 +1,1264 @@
...
+        return;
+    }
+    if (!repo) {
+        NEURO.client?.sendActionResult(actionData.id, true, 'You are not in a git repository.');
+        return;
+    }
+
+    const refToMerge = actionData.params?.ref_to_merge
+
+    if (!repo.getBranch(refToMerge)) {
```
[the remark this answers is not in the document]
KTrain5169: This is a bit hard then, since it's a check for if the branch exists. I think we just nuke this and just make it a check for if the field exists at all, pretty sure `repo.merge` should error if it doesn't exist anyways.

src/git.ts
```diff
@@ -739,7 +740,7 @@ export function handleGitMerge(actionData: any) {
         NEURO.client?.registerActions([
             {
                 name: 'abort_merge',
-                description: 'Aborts the current merge oepration.',
+                description: 'Aborts the current merge operation.',
                 schema: {}
             },
         ])
```
Pasu4, line 746: We currently register this action when there is any error, including errors that don't initiate a merge. We can check if there are conflicts if `repo.state.mergeChanges` is not empty.
Also, `abort_merge` should be unregistered if she resolves the conflict (I think there is no way to commit while there are merge errors so maybe we can unregister on commit?).

GitHub doesn't let me suggest a change, here is the code I suggest:
```ts
        if (repo?.state.mergeChanges.some(() => true)) {
            NEURO.client?.registerActions([
                {
                    name: 'abort_merge',
                    description: 'Aborts the current merge operation.',
                    schema: {}
                },
            ])
        }
```

src/git.ts
```diff
@@ -789,7 +790,7 @@ export function handleGitDiff(actionData: any) {
 
     const ref1: string = actionData.params?.ref1;
     const ref2: string = actionData.params?.ref2;
-    const filePath: string = actionData.params?.filePath;
+    const filePath: string = actionData.params?.filePath || repo.rootUri.fsPath;
```
Pasu4, line 793: Apparently git doesn't like absolute paths (?) 😓
`"."` does what we want.
```suggestion
    const filePath: string = actionData.params?.filePath || ".";
```

src/git.ts
```diff
@@ -0,0 +1,1219 @@
...
+        return;
+    }
+    if (!repo) {
+        NEURO.client?.sendActionResult(actionData.id, true, 'You are not in a git repository.');
+        return;
+    }
+
+    
+    const message = `Neuro commit: ${actionData.params?.message}`;
+    const commitOptions: string[] | undefined = actionData.params?.commitOptions;
```
Pasu4, line 668: Forgot about this in the last review.
```suggestion
    const commitOptions: string[] | undefined = actionData.params?.options;
```

src/git.ts
KTrain5169: inconsistency is fun

Pasu4 (approved):

PR #3, opened 2025-04-14
feat: mass-disable all neuropilot permission toggles

KTrain5169:
self-explanatory, disables all permission toggles immediately in the workspace and reloads permissions to ensure they are applied immediately

finishes one task of the v1.5.0 PR

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,6 +1,4 @@
 import * as vscode from 'vscode';
-import * as fs from "fs";
-import * as path from "path";
 
 import { NEURO } from './constants';
 import { logOutput, createClient, onClientConnected } from './utils';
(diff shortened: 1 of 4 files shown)


KTrain5169:
how the hell do I add reviewers

src/extension.ts
```diff
@@ -62,3 +63,22 @@ function registerPostActionHandler() {
...
+        }
+    }
+    Promise.all(promises).then(() => {
+
+        // Send context and reload
+        reloadPermissions();
+        NEURO.client?.sendContext("Vedal has turned off all dangerous permissions.");
+        vscode.window.showInformationMessage("All dangerous permissions have been turned off and actions have been re-registered.");
+    });
+}
```
Pasu4, line 84: This should also cancel Neuro's current task (and terminal for when that's implemented).

Pasu4 (changes requested):
A description of the command needs to be added to README.md. Also suggest the user add a keyboard shortcut for quick access.

src/extension.ts
KTrain5169: @Pasu4 merge this first, then I'll work on that after terminal is implemented.

PR #4, opened 2025-04-14
Refactor action handlers

Pasu4:
Refactors all action handlers so instead of sending an action result directly, they return an object that gets converted and sent in the parent handler. This should make it impossible to accidentally send action results twice or asynchronously.

~~Blocked by #5~~


KTrain5169:
can we delay this until my other PR is merged, then sync this branch with the `dev` branch so we can pull in teh changes? this way we can see the full list of handlers we need to modify.

PR #5, opened 2025-04-15
feat: terminal access for neuro

KTrain5169:
it's time for neuro to be given more power. to the command-line we go 😎

Neuro will be able to access any terminal defined in extension settings. By default:
```jsonc
[
    {
        "name": "PowerShell",
        "path": "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe",
        "args": []
    },
    {
        "name": "Command Prompt",
        "path": "C:\\Windows\\System32\\cmd.exe",
        "args": ["/K"] // Required to show anything in UI
    },
    {
        "name": "Git Bash",
        "path": "C:\\Program Files\\Git\\bin\\bash.exe",
        "args": []
    }
]
```

Upon Neuro selecting a shell and command to run, a Pseudoterminal will be spawned with the corresponding shell process. All stdout/stderr will be sent to Neuro to emulate a human looking through the terminal (instead of only sending command's stdout/stderr). Neuro can *also* kill shells by herself and see what shells are being ran.

These terminals will come up with their own tabs in the VSC UI and only one can ever be present. The time will also be displayed when a terminal starts but this is toggleable.
They will also use the following icon (designed by @Pasu4):
NeuroPilot Terminal Icon

For safety reasons, this is also hooked into the Disable All Permissions panic button to immediately kill all shell processes in addition to disabling the permission.


Pasu4:
Since this is supposed to be used on stream, it is more important that the chat can see what Neuro is doing, so we should go with the pseudoterminal approach.

To address terminal spam, we could let the user define what profile (or how many at once) Neuro is allowed to launch. Though we shouldn't implement this until we're sure terminal spam is a problem.

KTrain5169:
> Since this is supposed to be used on stream, it is more important that the chat can see what Neuro is doing, so we should go with the pseudoterminal approach.
> 
> To address terminal spam, we could let the user define what profile (or how many at once) Neuro is allowed to launch. Though we shouldn't implement this until we're sure terminal spam is a problem.

I'll still implement a terminal limit but probably will set it to something like 5. (Neuro shouldn't be using any more than that amount anyways)

KTrain5169:
> I'll still implement a terminal limit but probably will set it to something like 5. (Neuro shouldn't be using any more than that amount anyways)

ok so I forgot that the current implementation has a soft limit on terminals able to be open at once anyways so this isn't really necessary

Pasu4 (changes requested):

PR #6, opened 2025-04-29
refactor config gets + add toggle to allow neuro to see some unsafe paths

KTrain5169:
basically the same thing as the permissions getter but more adapted for configs

also be very careful with this as neuro can also obtain env vars


KTrain5169:
wtf is wrong with you vsc

image

PR #7, opened 2025-04-30
Improve editing

Pasu4:
## Changes

- Return context as a single code block, denote cursor position with `<<<|>>>`
  - ~~TODO: Test if Jippity can work with this syntax~~ Jippity doesn't work at the moment
- Return context for all editing actions
- Add an option to find / replace / delete with RegEx and replace with RegEx substitution
  - Supports both JS and .NET substitution syntax (except `$'` and `` $` `` because those would basically duplicate the file multiple times)
- Add an option to find / replace / delete first in file, last in file, first after cursor, last before cursor, all in file.


Pasu4:
Tested with ChatGPT: https://chatgpt.com/share/681a148e-c088-8004-8fcf-8853338ec2ad

PR #35, opened 2025-05-19
allow inline completions to be disabled

KTrain5169:
also refactor every instance of the err variable used to catch errors to erm because you can guess why

### DIFF
--- a/src/completions.ts
+++ b/src/completions.ts
@@ -7,13 +7,19 @@ import { CONFIG } from './config';
 let lastSuggestions: string[] = [];
 
 export function requestCompletion(beforeContext: string, afterContext: string, fileName: string, language: string, maxCount: number) {
-    if(!NEURO.connected) {
+    if (!NEURO.connected) {
         logOutput('ERROR', 'Attempted to request completion while disconnected');
         vscode.window.showErrorMessage('Not connected to Neuro API.');
         return;
     }
-    if(NEURO.waiting) {
+    if (NEURO.waiting) {
         logOutput('WARNING', 'Attempted to request completion while waiting for response');
+        return;
+    }
+
+    // If completions are disabled, notify and return early.
+    if (CONFIG.completionTrigger === 'off') {
+        vscode.window.showInformationMessage('Inline completions are disabled (completionTrigger is set to "off").');
         return;
     }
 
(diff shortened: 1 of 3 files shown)


src/completions.ts
```diff
@@ -7,16 +7,22 @@ import { CONFIG } from './config';
...
         logOutput('WARNING', 'Attempted to request completion while waiting for response');
         return;
     }
 
+    // If completions are disabled, notify and return early.
+    if (CONFIG.completionTrigger === 'off') {
+        vscode.window.showInformationMessage('Inline completions are disabled (completionTrigger is set to "off").');
+        return;
+    }
+
```
Pasu4, line 25: This should not show an information message, in case the user wants to use a different completion provider. Also, it should be moved above the `NEURO.connected` check, as showing that errors doesn't make sense when the feature isn't turned on.

Pasu4 (changes requested):

KTrain5169:
👍simple fix, I'll push it later

While we get this done should I also resolve the issue of allowing Neuro to fix problems?

I don't know if this is considered as a completion feature or if I throw the solution into another file.

src/completions.ts
```diff
@@ -7,12 +7,21 @@ import { CONFIG } from './config';
 let lastSuggestions: string[] = [];
 
 export function requestCompletion(beforeContext: string, afterContext: string, fileName: string, language: string, maxCount: number) {
-    if(!NEURO.connected) {
+    // If completions are disabled, notify and return early.
+    if (CONFIG.completionTrigger === 'off') {
+        vscode.window.showInformationMessage('Inline completions with NeuroPilot are disabled.');
```
Pasu4, line 12: If I turn the setting off, this pops up every time I try to use inline completions. It should at most pop up once per session or not at all.

Pasu4 (changes requested):

src/completions.ts
KTrain5169: I disagree since the user may forget about this setting being off and wondering why it's not working.

Pasu4: Constantly taking up a notification slot is not good. Can we find a better solution to this?

KTrain5169: We just timer it like RCE 🤣

src/completions.ts
```diff
@@ -7,12 +7,29 @@ import { CONFIG } from './config';
...
-    if(!NEURO.connected) {
+    // If completions are disabled, notify and return early.
+    if (CONFIG.completionTrigger === 'off') {
+        if (NEURO.warnOnCompletionsOff) {
+            return;
+        }
+        vscode.window.showInformationMessage('Inline completions with NeuroPilot are disabled.', 'Don\'t show again this session')
+            .then(selection => {
+                if (selection === 'Don\'t show again this session') {
+                    NEURO.warnOnCompletionsOff = false
```
Pasu4, line 18:
```suggestion
                    NEURO.warnOnCompletionsOff = false;
```

src/completions.ts
```diff
@@ -7,12 +7,29 @@ import { CONFIG } from './config';
 let lastSuggestions: string[] = [];
 
 export function requestCompletion(beforeContext: string, afterContext: string, fileName: string, language: string, maxCount: number) {
-    if(!NEURO.connected) {
+    // If completions are disabled, notify and return early.
+    if (CONFIG.completionTrigger === 'off') {
+        if (NEURO.warnOnCompletionsOff) {
```
Pasu4, line 12:
```suggestion
        if (!NEURO.warnOnCompletionsOff) {
```

Pasu4 (approved):

PR #40, opened 2025-05-21
feat: allow neuro to explain & fix linting errors, like copilot

### LINKED ISSUE #25: Let Neuro fix linting problems
Don't know if this is possible, but would likely require proposed API. When you press `Ctrl+.` on a linting problem, there is an option to let copilot fix it, what we want is to let Neuro fix it instead. This shouldn't require Autopilot, since GitHub Copilot can do this too, there must some syntax to this.

KTrain5169:
I don't know what I'm doing so I'll just open this draft now and maybe deal with it later

resolves #25 

This adds to the Copilot side of NeuroPilot by adding the `Ask Neuro to fix` and `Ask Neuro to explain` buttons to lint errors.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,7 +1,7 @@
 import * as vscode from 'vscode';
 
 import { NEURO } from './constants';
-import { logOutput, createClient, onClientConnected } from './utils';
+import { logOutput, createClient, onClientConnected, isPathNeuroSafe, setVirtualCursor } from './utils';
 import { completionsProvider, registerCompletionResultHandler } from './completions';
 import { giveCookie, registerRequestCookieAction, registerRequestCookieHandler, sendCurrentFile } from './context';
 import { registerChatParticipant, registerChatResponseHandler } from './chat';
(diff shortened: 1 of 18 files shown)


src/extension.ts
```diff
@@ -35,6 +37,143 @@ export function activate(context: vscode.ExtensionContext) {
...
+        const response = await requestAIResponseForDiagnostic(document, diagnostic, 'explain', tokenSource.token);
+
+        // Append the response to the chat history.
+        addToChatHistory("Explain: " + response);
+        // Show an info message with an option to view the chat history.
+        const choice = await vscode.window.showInformationMessage("Neuro explains: " + response, "View Chat History");
+        if (choice === "View Chat History") {
+            chatHistoryChannel.show();
+        }
+    });
```
Pasu4, line 114: This takes up too much space in `activate`, the lambdas should be extracted into functions (lint_problems.ts?)

src/extension.ts
```diff
@@ -35,6 +37,143 @@ export function activate(context: vscode.ExtensionContext) {
...
+                actions.push(explain);
+            });
+
+            return actions;
+            }
+        })(),
+        {
+            providedCodeActionKinds: [vscode.CodeActionKind.QuickFix]
+        }
+    );
```
Pasu4, line 174: I don't quite understand what this does at the moment (1:30 AM brain), but if possible extract from `activate`.

Pasu4 (changes requested):
Didn't do any testing, just looked over the code real quick, I'll do a proper review tomorrow.

Pasu4 (approved):

PR #41, opened 2025-05-21
Implement Neuro cursor

### LINKED ISSUE #31: Give Neuro her own cursor
Currently, Neuro uses the actual cursor for editing. This becomes a problem when using editing actions in Copilot Mode, since the user might move the cursor away from where Neuro wants to edit.

Instead, we should use a virtual cursor that is used for Neuro's editing actions. To make it visible to the user (and the viewers), we can use Text Editor Decorations (a working example below). It should only be displayed if editing has at least Copilot permission level. There should also be a setting to make the real cursor follow Neuro's cursor.

```ts
const deco = vscode.window.createTextEditorDecorationType({
    backgroundColor: 'rgba(0, 0, 0, 0)',
    border: '1px solid rgba(0, 0, 0, 0)',
    borderRadius: '1px',
    overviewRulerColor: 'rgba(255, 85, 229, 0.5)',
    overviewRulerLane: vscode.OverviewRulerLane.Right,
    gutterIconPath: context.asAbsolutePath('icon.png'),
    gutterIconSize: 'contain',
    rangeBehavior: vscode.DecorationRangeBehavior.ClosedClosed,
    before: {
        contentText: 'ᛙ',
        margin: '0 0 0 -0.25ch',
        textDecoration: 'none; position: absolute; display: inline-block; top: 0; font-size: 200%; font-weight: bold, z-index: 1',
        color: 'rgba(255, 85, 229)',
    },
});

vscode.window.onDidChangeActiveTextEditor(editor => {
    if (editor) {
        editor.setDecorations(deco, [{range: new vscode.Range(1, 4, 1, 4), hoverMessage: 'Neuro'} satisfies vscode.DecorationOptions]);
    }
});
```
`'ᛙ'` (RUNIC LETTER SHORT-TWIG-MADR M) is the symbol that Live Share also uses for their cursors.

Pasu4:
Resolves #31

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,11 @@ Allows the user to specify the flavor as well.
 Attempts to reconnect to the API.
 Shows a notification when it succeeds or fails.
 
+### Move Neuro's Cursor Here
+
+Moves Neuro's cursor to the current cursor position.
+Only works if there is an active text editor, and only in files Neuro has access to.
+
 ### Reload Permissions
 
 Reregisters all actions according to the permissions.
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -37,6 +37,7 @@ export default tseslint.config(
             '@stylistic/no-trailing-spaces': ['warn', { 'ignoreComments': true }],
             '@stylistic/quotes': ['error', 'single', { 'avoidEscape': true }],
             '@typescript-eslint/no-empty-function': 'off',
+            'prefer-const': 'warn',
             '@typescript-eslint/naming-convention': [
                 'warn',
                 {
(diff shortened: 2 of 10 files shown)


KTrain5169 (approved):
is there a reason why we don't keep track of the cursor position in the undo stack? otherwise looks good at first glance

Pasu4:
> is there a reason why we don't keep track of the cursor position in the undo stack?

The API doesn't expose the undo stack. Also undoing in general doesn't work that well if multiple people work in the same file, even live share has this problem.

> otherwise looks good at first glance

There is still probably some errors here, for example the position of the cursor does not update if the user types something before it.

PR #47, opened 2025-05-27
Add autopilot gif, remove old gif

Pasu4:


KTrain5169 (approved):
could also be good to show what happens using Tony for stuff like receiving save/lint notifications but this is enough as a basic thing

Pasu4:
> could also be good to show what happens using Tony for stuff like receiving save/lint notifications but this is enough as a basic thing

I think this can be a screenshot, I'll do that later.

PR #59, opened 2025-06-17
assets page

KTrain5169:
I think I have to play catch-up with this later


Pasu4:
Currently missing the new icon by Xaendril (assets/heart-xaendril.png)

KTrain5169:
I was waiting on it before but apparently GIMP won't let them export to SVG so

Issue #32, opened 2025-05-18
Lighter version for browser/remote use?


KTrain5169:
I don't know if this is even possible, but I'll open an issue anyways in case we can.

For a while now (when making autopilot features) we've been operating on the assumption that Vedal/whoever will run this in a VM. But until now, we're not exactly making that easy.

Therefore, a possible separate version (or maybe not we could try and integrate it into this extension if we want) used for remote connections should be considered. This would probably not be that hard once we get an initial base working. This would limit features to be able to what is available in the limited environment of the VS Code Web browser at least.

The biggest reason against this would be it being too much work (2 codebases, or an even bigger codebase) just for something that could be potentially mitigated using other tools.

Pasu4:
Are there problems with running in the browser at the moment?

KTrain5169:
yes, in that you cannot even run it in the browser due to not being marked as supported. And since the browser is restricted to the WebWorker environment not everything in the extension can run.

WebSockets and non-file-system related stuff work, Git is conditional apparently (github.dev activates the Git extension, but otherwise none).

Additionally, I tested NeuroPilot with an SSH connection to a local VM. Because NeuroPilot is required to run in the Extension Host it gets installed on the SSH server, which would definitely not be where Neuro is installed.

Luckily for the browser environment, you can specify an entirely different entrypoint for it, meaning that if we use, say, `./out/extension-web.js` as its entrypoint, we can restrict the set of runnable actions to be smaller.

As for Remote Development extensions such as SSH, WSL, etc... there is a way to specify it to be installed on the client environment, but things will be ran on the local client instead, which may not be what we want.

KTrain5169:
(this only applies to browser version, remote versions are still being figured out)

Doing a bit of research on this issue while I can atm (don't have access to my usual tools right now so I can't work on any of the actual important issues + I can't even remember them off the top of my head), I've ruled out the pseudoterminals, tasks, and potentially even chat & completions actions as unable to run in the web. File actions doesn't need to be reworked (more or less) to make it work in the browser, and git actions can work but only because we have a gitValidator now that checks if the git extension is even activated, in addition to Commit becoming Commit & Push.

Additionally we may have to rewrite some code so it uses WebWorkers. We also need to make sure the browser entrypoint is one singular file because that's the restriction apparently.

[...]

PR #65, opened 2025-06-20
feat: web build for neuropilot

### LINKED ISSUE #56: Bundling extension output
This would allow NeuroPilot to load faster and would also be more convenient for devs since we don't have to include node_modules which might accidentally contain unnecessarily bundled modules (somewhat great example would be the types since we're just running pure JS)

Also would avoid having to wrangle 2 different compilers together for web build (ik webpack is very old, but when I was testing it, it would have some error that I think is due to `/out` already existing, not sure if esbuild will have that same issue)

### LINKED ISSUE #69: Add preselect dropdown menu for docs page
Prep for #32 / NeuroPilot API milestone

Companion extensions hooking in should be able to register their docs links, which should then be added to the list of docs sites you can choose to view when you click "View Docs Homepage" or smth like that

and yes I'm calling them companions, kinda fits with the vibe

KTrain5169:
Still trying to figure out some issues that are occurring with regards to runtime but we'll see

Also I have not modified CI yet

We're bundling the extension outputs now as well, so #56 is resolved here


src/extension.ts
```diff
@@ -12,6 +12,19 @@ import { CONFIG } from './config';
 import { explainWithNeuro, fixWithNeuro, NeuroCodeActionsProvider, sendDiagnosticsDiff } from './lint_problems';
 import { editorChangeHandler, fileSaveListener, moveNeuroCursorHere, toggleSaveAction, workspaceEditHandler } from './editing';
 import { emergencyDenyRequests, acceptRceRequest, denyRceRequest, revealRceNotification } from './rce';
+import { GitExtension } from './types/git';
+import { getGitExtension } from './git';
+
+const docsOptions: Record<string, string> = {
```
KTrain5169, line 18: This is not working actually, idk why.

package.json
```diff
@@ -138,6 +139,11 @@
 				"title": "Switch Neuro API user name",
 				"category": "NeuroPilot"
 			},
+			{
+				"command": "neuropilot.refreshExtensionDependencyState",
+				"title": "Get dependency extension states",
```
Pasu4, line 144: Title doesn't match what it's doing. Also mention the upcoming deprecation of this command.
```suggestion
				"title": "Refresh extension dependencies (workaround, will be deprecated soon)",
```

src/extension.ts
```diff
@@ -35,38 +49,81 @@ export function activate(context: vscode.ExtensionContext) {
...
+                `${selectedOption} Docs (WebView)`,
+                vscode.ViewColumn.Active,
+                { enableScripts: true },
+            );
+            panel.webview.html = openDocsPage(baseUrl);
+            quickPick.hide();
+        });
+
+        quickPick.show();
     });
```
Pasu4, line 81: Extract to function to keep `activate` clean.

src/extension.ts
```diff
@@ -35,38 +49,81 @@ export function activate(context: vscode.ExtensionContext) {
...
 
     vscode.commands.registerCommand('neuropilot.openSpecificDocsPage', async (args?: { subpage?: string }) => {
```
Pasu4, line 83: Extract to function to keep `activate` clean.

Pasu4 (changes requested):
(As said in the discord) There are two duplicated files with minor differences in the `src/web` folder. If at all possible, I would like them to be "de-duplicated", e.g. by putting all code that is the same in both files into one file and have only the differences in two separate files. That way, we can avoid situations like the `docsOptions` problem a few days ago.

src/extension.ts
KTrain5169: Yeah I figured you can't.

Pasu4 (changes requested):
For extension.ts and pseudoterminal.ts, I can't see well what exactly changed, so I'm just going to trust you that they're functionally equivalent.

Also, please don't commit more changes unrelated to #32 to this branch, as it makes it harder to review (except for requested changes of course).

.github/workflows/docs.yml
```diff
@@ -80,12 +80,20 @@ jobs:
               id: pages
               uses: actions/configure-pages@v5
             
+            - name: Get version
+              id: docs-ver
+              uses: ./.github/actions/get-package-json-version
+              with:
+                path: ./docs
+
```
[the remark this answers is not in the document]
KTrain5169: nope, thought I got rid of it

.github/actions/get-package-json-version/dist/index.js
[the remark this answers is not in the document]
KTrain5169: Actions require this since they aren't built on-demand (well they can be but that deals with docker)

KTrain5169:
> For extension.ts and pseudoterminal.ts, I can't see well what exactly changed, so I'm just going to trust you that they're functionally equivalent.

I had to reshuffle the locations of some things as I was concerned that esbuild/tsc would pick it up and bundle/warn

> Also, please don't commit more changes unrelated to #32 to this branch, as it makes it harder to review (except for requested changes of course).

👌

Pasu4 (approved):

PR #68, opened 2025-06-23
Unit tests

### LINKED ISSUE #12: Get unit tests to work
The reason why NeuroPilot currently has no unit tests is because I couldn't get the VS Code to even pass the `1 === 1` check, it would always crash before getting to that point. There should be unit tests for most of the utils methods at least. Info on testing extensions: https://code.visualstudio.com/api/working-with-extensions/testing-extension
Also there should be CI that warns if any tests fail.

### LINKED ISSUE #13: Add integration tests
~~*Blocked by #12*~~

Currently all testing is done manually, which is a huge time sink for development and also reviews. This should be automated so that it can be more easily verified that a change doesn't break anything. Ideally, we use VS Code's integration testing framework for this (if we get it to work — this depends on #12). An alternative would be to write the tests in Python and use Tony's API, though I would rather avoid that.

The following tests will be / have been created:

- [ ] chat.test.ts
- [ ] completions.test.ts
- [ ] config.test.ts
- [ ] constants.test.ts
- [ ] context.test.ts
- [ ] editing.test.ts
- [ ] extension.test.ts
- [x] file_actions.test.ts
- [ ] git.test.ts
- [ ] lint_problems.test.ts
- [ ] neuro_client_helper.test.ts
- [ ] pseudoterminal.test.ts
- [ ] rce.test.ts
- [ ] tasks.test.ts
- [ ] unsupervised.test.ts
- [ ] utils.test.ts

*(This checklist was directly copied from #68 and may require some revision with the new changes from #108)*

Pasu4:
~~Resolves #12, resolves #13~~

Tests use mocha and ts-mockito.

The following tests will be / have been created:

- [ ] chat.test.ts
- [ ] completions.test.ts
- [ ] config.test.ts
- [ ] constants.test.ts
- [ ] context.test.ts
- [ ] editing.test.ts
- [ ] extension.test.ts
- [x] file_actions.test.ts
- [ ] git.test.ts
- [ ] lint_problems.test.ts
- [ ] neuro_client_helper.test.ts
- [ ] pseudoterminal.test.ts
- [ ] rce.test.ts
- [ ] tasks.test.ts
- [ ] unsupervised.test.ts
- [ ] utils.test.ts

~~Some of these may be actually untestable, e.g. because I can't check if a popup window is being shown (or maybe I can, we'll see).~~ Pretty much everything is testable with ts-mockito.


.vscode-test.mjs
```diff
@@ -0,0 +1,6 @@
+import { defineConfig } from '@vscode/test-cli';
+
+export default defineConfig({
```
KTrain5169, line 3:
```suggestion
export default defineConfig({
    label: 'desktopUnitTest'
```

I need to figure a better way to do this

KTrain5169:
this technically falls under refactor and docs based on label desc btw

KTrain5169:
@Pasu4 Is it possible to now look at maybe hosting the rewire page outputs to another repository now that we are in an org? This way it'll show under (example) `https://vsc-neuropilot.github.io/unit-tests` instead of having a sorta hacky workaround.

Pasu4:
> this technically falls under refactor and docs based on label desc btw

It's not supposed to, refactor is for improving code that already exists. I'll make the label description more clear.

> @Pasu4 Is it possible to now look at maybe hosting the rewire page outputs to another repository now that we are in an org? This way it'll show under (example) `https://vsc-neuropilot.github.io/unit-tests` instead of having a sorta hacky workaround.

We can do that (though this isn't directly related to rewire, mocha outputs the pages I think), but I would say `/neuropilot-unit-tests` in case we want to make unit tests for other extensions. Or `/unit-tests/neuropilot` if possible.

KTrain5169:
I think `/unit-tests/neuropilot` is possible, I can go set up a repository for that really quickly

We will be setting up more unit tests because of the neuropilot-git and neuropilot-server splits so having it all be under /unit-tests makes sense

KTrain5169:
I'm assuming you want to not keep this for long? Unless you don't count on rewire maintainers responding anytime soon.

KTrain5169:
Also, should unit tests be done on multiple platforms (windows, macos, ubuntu) or just windows/ubuntu?

Prepping the unit test repository right now including all the moving parts necessary

KTrain5169:
> Yes. Someone set this up with 'matrix' for Tony's CI at some point, so you can look there maybe.

I was already thinking about doing this anyways + unit testing web build on firefox + webkit (safari) + chromium so I already mapped out a way to do this.

KTrain5169:
> We can do that (though this isn't directly related to rewire, mocha outputs the pages I think), but I would say `/neuropilot-unit-tests` in case we want to make unit tests for other extensions. Or `/unit-tests/neuropilot` if possible.

I've set up `VSC-NeuroPilot/{unit-tests,actions}` to do this (mostly), will need to figure out how to rig up CI to do this.

.vscodeignore
```diff
@@ -20,4 +20,7 @@ tony/
 # Docs
 docs/**
 !docs/public/*.gif
-!docs/public/*.png
\ No newline at end of file
+!docs/public/*.png
+
+# Coverage
+/coverage
```
KTrain5169, line 26: Also ignore test-related files.

```suggestion
/coverage

# Integration testing
test-playground/
.vscode-test.mjs
```

(src/test/** is already ignored afair)

test-playground/.vscode/tasks.json
```diff
@@ -0,0 +1,12 @@
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "Run Current File",
```
KTrain5169, line 5: Unless I'm missing something (I just skimmed through really quickly and haven't taken a closer look) shouldn't we have a .js file if we are using this task?

KTrain5169 (changes requested):
preliminary review

.vscodeignore
Pasu4: `test-playground` should not be ignored because the `.vscode/settings.json` in it are required for the tests to run in Autopilot mode.

src/test/suite/index.ts
```diff
@@ -0,0 +1,38 @@
...
+                    });
+                } catch (erm) {
+                    e(erm);
+                }
+            })
+            .catch(erm => {
+                return e(erm);
+            });
+    });
+    return new Promise((_resolve, _reject) => {});
```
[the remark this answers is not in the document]
Pasu4: I just copied that from the guide, idk why it's there

.vscodeignore
Pasu4: I'm not sure what you mean, the folder must exist for it to be opened, no?

Pasu4: I see. 💀

src/config.ts
```diff
@@ -22,7 +22,8 @@ export function get<T>(key: string): T | undefined {
  *
  * @param permissions The permission(s) to query.
  * @returns The lowest permission level in the list of permissions.
- * If no permissions are specified, this function assumes Copilot
+ * If no permissions are specified, this function assumes {@link PermissionLevel.COPILOT}.
+ * If used as a boolean, {@link PermissionLevel.OFF} is considered `false`, everything else is considered `true`.
```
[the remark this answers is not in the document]
Pasu4: We're really using enum values that can be implicitly cast to booleans. `OFF` has a value of zero, which is considered `false` if used as a boolean, `COPILOT` and `AUTOPILOT` are 1 and 2 respectively, which are both considered `true` if used as a boolean.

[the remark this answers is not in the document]
KTrain5169: hmmm

Issue #73, opened 2025-07-11
Delete a good portion of the README and replace with docs mentions


KTrain5169:
The README is getting pretty big and if we don't do something about it it's just gonna be an endless pit of annoyance.

PR #76, opened 2025-07-17
documentation updates

KTrain5169:


docs/src/content/docs/contributors/meta.mdx
```diff
@@ -0,0 +1,70 @@
...
+---
+
+import { Aside } from '@astrojs/starlight/components'
+
+## CI (GitHub Actions) :badge["External"]
+
+### Workflows
+
+- Changing CI shouldn't be done often, it should only be changed if something within the project requires it.
+- Enabling GitHub Actions on forks is recommended so that you can push your changes and get feedback, however you should disable the publishing workflow on your fork.
```
Pasu4, line 14: Maybe we should build in a check that skips/fails the workflow if it's not on in the right repo, to prevent people accidentally publishing (which I assume this is for).

docs/src/content/docs/reference/cursor.mdx
```diff
@@ -0,0 +1,19 @@
...
+Neuro's virtual cursor is a pink cursor decoration that is visible whenever Neuro has any level of access to [editing the active document](/neuropilot/reference/permissions#edit-active-document).
+This virtual cursor is controlled almost entirely by Neuro, and is used in place of the real cursor which the user can control independently.
+
+This cursor is important as it allows separation of cursors in actions such as inserting text. This allows the user to move the cursor while still benefitting from Neuro asking to insert text/look around her cursor position based on where she wants to look.
+
+The user can manually move Neuro's cursor using the [Move Neuro's Cursor Here](/neuropilot/reference/commands#move-neuros-cursor-here), and can also let her sync the real cursor with the virtual cursor using the [Cursor Follows Neuro](/neuropilot/reference/settings#cursor-follows-neuro) setting.
+
+<Aside>
+
+The cursor will not appear while Edit Active Document is set to 'Off'.
```
Pasu4, line 17: Add the other reason why the cursor might not appear.
```suggestion
The cursor will not appear while Edit Active Document is set to 'Off' or in documents Neuro is not allowed to edit.
```

docs/src/content/docs/reference/dependencies.mdx
```diff
@@ -9,7 +9,7 @@ Multiple features require another extension to work, so this page lists them so
 
 <Aside type="tip">
 
-You probably will most likely only needing to be looking for extension dependencies, as the dependencies from `npm` that we use should already be bundled in the extension pack.
+You probably will most likely only needing to be looking for extension dependencies, as the dependencies from `npm` that we use should already be bundled in the extension build.
```
Pasu4, line 12: Grammar
```suggestion
You will most likely only need to look for extension dependencies, as the dependencies from `npm` that we use should already be bundled in the extension build.
```

docs/src/utils/version.ts
```diff
@@ -0,0 +1 @@
+export const docsVersion: string | undefined = import.meta.env.PUBLIC_VERSION
```
Pasu4, line 1: Does this work? I think what sets this variable is still in the other PR (also there's a shell command that gets the version, `npm show ./ version` iirc., so do we need a whole action for that?).

docs/src/content/docs/contributors/meta.mdx
KTrain5169: They can't publish it without intentionally changing the publisher name (still currently under your name) & adding their own marketplace token, at which point, we can't do anything.

Also the publishing workflow runs on dispatch anyways so it doesn't run on PRs, I completely forgot about that. I guess it's more as a precaution at this point.

KTrain5169:
Also got #73 out of the way because I figured why not do it now

PR #84, opened 2025-08-07
alias imports via tsconfig.json

KTrain5169:
alias'ed imports:

- ~/* = src/*
- @typing/* = src/types/*
- @shared/* = src/shared/*

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -2,10 +2,10 @@ import * as vscode from 'vscode';
 import { NeuroClient } from 'neuro-game-sdk';
 import globToRegExp from 'glob-to-regexp';
 
-import { NEURO } from './constants';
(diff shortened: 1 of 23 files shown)


src/utils.ts
```diff
@@ -228,7 +228,7 @@ export function escapeRegExp(string: string): string {
  * @throws Error if the substitution is invalid or if the capture group does not exist.
  */
 export function substituteMatch(match: RegExpExecArray, replacement: string): string {
-    const rx = /\$<.+?>|\${.+?}|\$\d+|\$./g;
+    const rx = /\$<.+?>|\${.+?}|\$\d+|\$~/g;
```
Pasu4, line 231: This was probably an oversight while replacing all occurrences.

```suggestion
    const rx = /\$<.+?>|\${.+?}|\$\d+|\$./g;
```

Issue #51, opened 2025-06-06
.neuropilot/


KTrain5169:
Towards the next release (or the release after) I think one of the important things to consider is the `.neuropilot` folder. Right now we are currently relying on the .vscode/settings.json to store our configs. Not a bad measure by any means, but we could do better by moving it to our own folder.

Things to consider:
- What should the schema for the JSON look like?
- What items should be moved to the .neuropilot folder? (Permissions is a definite "no")
- Should we make our own UI for our settings?

PR #88, opened 2025-08-26
Patches PR following the 2025-08-25 Evil Dev Stream test!

KTrain5169:
# @KTrain5169's list:
## Functional changes

- [x] When switching files by a human, context isn't sent, this is a bit of a problem for "student-pilot" as vedal switching files may not be made clear to the twins & they don't get a refresh on the file's contents *(Implemented in fb4aa2cd0074d66b1d588bd10fbc5d8c7698a658)*
- [x] Evil tries to input position parameters in insert text instead of moving the cursor first and then inserting text, this should be allowed and all schemas **must** now have `additionalProperties: false` for the schema lib (unless we just add a custom check for that at RCE instead of back-adding that to every schema) *(Implemented in b942e11dc9453ebb7bfe84015ef92d57249c06a6 (position parameters) and 8354ef6edd5307c372dd327b2d837443de1a4f03 (additionalProperties))*
- [x] A different option for regex text should be available as an option for replace/delete - Glob patterns probably weren't tried during stream. Probably should've realised that much earlier but oh well.
- [x] Range options should be specifiable for replace/delete *(Implemented in b0d0cb801f3c89b04fad1f482c4d750702ceb5e7)*
  - This should be used to limit the range to replace and delete text.
- [x] actually while we're at it the cursor should also have the ability to highlight like clicking and dragging text like a range *(Implemented in df8dbbb92c3f415e98a64574d547de0bbd1b71f1)*

## Config/context changes

- [x] CNAME shouldn't be neuro-safe by default (currently is because it doesn't start with a dot and lives at root, not a subfolder)
- [x] perhaps should be made clear to evil that binary files like .png are gibberish (is there even a way to tell if a file is binary without trying to parse it?) *(Implemented in 8e812ad503c8cb063944ccf92c664889ab7d7834)*

# @Pasu4's list:
## General changes

- [x] Specify to use newlines and indentation appropriately for `insert_text`. *(Implemented in 5a8cd5ddbcbc4788f9bcf347a0a309bd57884c63)*
- [x] The user should be able to disable specific actions by name. *(Implemented in c2644db7fb19d5ab4c2d6cdacf1e9438fe649bd8)*
  - ~~VS Code settings UI might get a bit messy if this was implemented, so something like #51 and/or a sidebar to contain toggles should be used here instead.~~ @KTrain5169 learned something new and that was array schemas for extension settings.
- [x] The entire file should be sent as context. *(Kinda implemented in dbe0f3d7f00b1b87f4dc3fa8714e2b598e15ea67)*
  - Maybe a max length failsafe would be good in case she does something with regex replace again, but what should happen in that case?
- [x] If an action moves the cursor, it should be mentioned in its description. *(Implemented in 8151bb7061d5b79a33505e43bba2a71502b3b407)*
- [x] Show Neuro's last edit using text decorations. *(Implemented in 41f8f67592f3f396c70287cda2ae68df3d5728f0)*
  - We could use [jonTrent/PatienceDiff](https://github.com/jonTrent/PatienceDiff) for this (licensed under [The Unlicense](https://github.com/jonTrent/PatienceDiff/blob/dev/UNLICENSE.txt))

## New actions

- [x] `get_content`: Get the content of the current file. *(Implemented in a2f4fccde50470c2266fb239c2e50931a758e8fa)*
  - Under the hood, this should just execute the `Send Current File as Context` command, or use the same code.
- [x] `insert_lines`: Insert lines of code below the specified line. *(Fully implemented in 269136b36b448c47c6296496f8ba71031059eac4)*
  - Would this be using arrays or `\n` characters in the action object?
- [x] `rewrite_lines`: Rewrite everything in the specified line range. *(Implemented in #94)*
- [x] `delete_lines`: Delete everything in the specified line range. *(Implemented in #95)*
- [x] `rewrite_all`: Rewrite the entire contents of the file. *(Implemented in #93)*
- [x] `read_file`: Send the contents of a file as context without opening / switching to it. *(Implemented in 9229cb05cd0fdd3bdd2ec3f29e7ceb08c8e0b7a2)*
  - Evil wanted to "look at the placeholders for motivation while i do the css" at one point
  - Not sure if that's possible with the VS Code API but I'll (@Pasu4) look into it

# Issues to also resolve with this branch:

- [x] #89 - Add a changelog *(in progress by @KTrain5169, test implementation on the `dev` branch)*
  - We can use the [changesets](https://github.com/changesets/changesets) package for this.
    - See [this discussion](https://github.com/changesets/changesets/discussions/1297) for tips on how to set it up with VS Code extension development.


KTrain5169:
@Pasu4 I'm currently working on adding the position parameters for the editing actions. If you're going to continue making it (probably better rn), my suggestion is to refactor the schema object so that position parameters are deduplicated from the move cursor action and insert text action. Also, parameters should be a separate object inside the schema so that we can mark both line & column as required, but not if she doesn't input position params (unless we don't want to make both required at once and instead just default to cursor, I think I discussed this on Neurocord but I'm not entirely sure if that's a good idea now)

KTrain5169:
> Also, additionalProperties isn't recursive so you need to add it to the inner object (just saying this in case you didn't know).

No I remembered earlier today, but I did forget yesterday. I'll make sure to add that to my commits.

> I was more thinking of a string list setting of disabled action names, I don't think that would get messy. I'll try it tomorrow to see if it's really that bad.

String lists with \n are something I want to avoid simply because it's not typo-proof, at least with .neuropilot/ we can add a schema that says "hey by the way, this is a non-existent action" but I'll leave that up to you.Message ID: ***@***.***>strstrstrstrs

Pasu4:
> Would this be using arrays or `\n` characters in the action object?

`\n` would probably be better so she can write continuous code, which I assume would be easier for her.

KTrain5169:
> > Would this be using arrays or `\n` characters in the action object?
> 
> `\n` would probably be better so she can write continuous code, which I assume would be easier for her.

Looks like she's able to do continuous lines of code as well with \n so she definitely understands the assignment.

src/utils.ts
```diff
@@ -413,3 +413,13 @@ export function getProperty(obj: unknown, path: string): unknown {
 
     return current;
 }
+
+/**
+ * Checks if the extension is currently on a virtual file system.
+ */
+export function checkVirtualWorkspace(_actionData: ActionData): ActionValidationResult {
+    if (vscode.workspace.workspaceFolders?.every(f => f.uri.scheme !== 'file')) {
+        return actionValidationFailure('You are currently in a virtual workspace.');
```
Pasu4: This should also explain why this fails, instead of just stating that. Otherwise this might sound just like general information.
```suggestion
        return actionValidationFailure('You cannot perform this action in a virtual workspace.');
```

Pasu4: Wait why is this outdated

KTrain5169: weird ways to work git amirite...

KTrain5169:
@Pasu4 did a bit of looking for you regarding the read_file action, you can read a file, get its Uint8Array value, and parse it using `Buffer.from(fileData).toString('utf8');`. I've already done the binary file validator (which uses the same sort of logic but looks at the Uint8Array value directly), so you can use that to validate that she isn't going to get a load of binary rubbish first.

I don't think there's a simpler way unless the text editor is already open.

PR #94, opened 2025-08-29
Add rewrite_lines action and implement character limit validation

frogneko:
- Add rewrite_lines action to replace content in specific line ranges
- Implement 100,000 character limit validation for all string parameters in editing actions
- Create generic createStringValidator function for consistent parameter validation
- Add string validation to insert_text, insert_lines, replace_text, delete_text, find_text, rewrite_all, and rewrite_lines actions

### DIFF
--- a/src/editing.ts
+++ b/src/editing.ts
@@ -24,6 +24,32 @@ const POSITION_SCHEMA = {
 };
 
 /**
+ * Create a generic string validator that checks type and character limits.
+ * @param paramPaths Array of parameter paths to validate (e.g., ['text', 'content'])
+ * @param maxLength Maximum character length (default: 100,000)
+ * @returns A function that validates the specified string parameters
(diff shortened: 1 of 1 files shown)


frogneko:
> In this case it's ok, but in the future if you want to add a new feature that's not part of the current plans (like the character limit), it would be better to discuss this in the discord beforehand. I don't want you to spend your time on something which doesn't end up getting accepted.

Fair point! Since I was going to add it to my new action as a validator it made sense to add to any similar actions in the end but I'll bring it up in the discord first, for similar things in the future

frogneko:
> Just out of curiosity, what got you to land on the 100000 character limit as default?

Tried to figure out a very large limit, more of which would never be necessary, but something that wouldn't block any neuro work. It was still quite a fuzzy guess at best :)

PR #107, opened 2025-09-12
reload permissions on permission change automatically

KTrain5169:
also properly block workspace-level config changes to terminal access permission & initialises extension onStartupFinished

### DIFF
--- a/src/shared/extension.ts
+++ b/src/shared/extension.ts
@@ -70,7 +70,7 @@ export function setupCommonEventHandlers() {
                 setVirtualCursor();
             }
             if (event.affectsConfiguration('neuropilot.permission') || event.affectsConfiguration('neuropilot.disabledActions')) {
-                vscode.commands.executeCommand('neuropilot.disableAllPermissions');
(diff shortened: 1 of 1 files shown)


Pasu4:
> should `neuropilot.disabledActions` reload actions automatically?

Yes

package.json
```diff
@@ -720,7 +720,7 @@
 			"supported": "limited",
 			"description": "Terminals and Tasks access will be forcefully disabled for Neuro. She will also be unable to access Neuro-unsafe paths and won't have all tasks registered.",
 			"restrictedConfigurations": [
-				"neuropilot.permission.directTerminalAccess",
+				"neuropilot.permission.terminalAccess",
```
Pasu4, line 723: Should be added to the changelog under fixes. Something like *"Terminal access permission now cannot be enabled in untrusted workspaces (didn't work before because of wrong setting ID)"*. Avoid making is sound like Neuro had terminal access in untrusted workspaces before (e.g. **NOT** like *"Terminal access permission is now restricted in untrusted workspaces."*).

Pasu4 (approved):

PR #129, opened 2025-09-27
feat: api

KTrain5169:
long-running pr, definitely won't be merged any time soon.

initial version intentionally doesn't work because I don't think I know what I'm doing yet.

terminology:

- Companions: The name of the extension(s) that use our API
- yet to add more


KTrain5169:
@Pasu4 I think it's probably best that this PR be closed, the branch recreated and the proposal PR recreated as well, since this branch is way too out-of-date and the current "API" will need to be rewritten anyways. Thoughts?

PR #131, opened 2025-10-02
feat: experimental schema switch

### LINKED ISSUE #127: Use `oneOf` for schema `git_diff` schema validation
Since we're already using `additionalProperties` everywhere, can we also use `oneOf` for the complicated `git_diff` schema? Would alleviate having to have the really not necessary validator, but this means that if Neuro fills out an extra field it will be rejected instead of approved.

KTrain5169:
Closes #127 

`oneOf` checklist:

- [x] `git.ts` - git_diff action
- [x] ~~`editing.ts` - position params~~

`description`/`examples` checklist:

- [x] `editing.ts`
- [x] `file_actions.ts`
- [x] `git.ts`
- [x] `lint_problems.ts`
- [x] `pseudoterminal.ts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -962,7 +962,7 @@
 		"@types/node": "^20",
 		"@types/vscode": "^1.95.0",
 		"@types/ws": "^8.18.0",
-		"@vscode/test-web": "^0.0.73",
+		"@vscode/test-web": "^0.0.74",
 		"@vscode/vsce": "^3.6.2",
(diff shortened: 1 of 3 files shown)


Pasu4 (changes requested):
Just as a fail-safe, it would be good to have a setting to remove experimental keys from the schemas sent to Neuro, just in case it somehow fails catastrophically. We already have a filtering function that we can expand with this functionality.

PR #142, opened 2025-10-18
Wrap all of Neuro's actions inside a try-catch

### LINKED ISSUE #139: Wrap actions in a try-catch and surface an error popup with instructions on how to report.
Needs to be extended with the API later to allow extensions to add custom issue report links for their actions.

This should encompass all our actions, including `request_cookie`.

Could also be extended to VS Code commands as well maybe, idk :shrug:

KTrain5169:
Need some opinions on the disable action button, namely:

- Should we add more disable options?
- Should we auto-detect the scope in which the action is enabled?
- Should we make the disable button only disable for the session?

Closes #139

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,8 @@ Changes between each version before then will not be listed.
 
 - The `diff_patch` action had an incomplete example. This has now been fixed.
 - The workspace lint validator wasn't implemented properly and would skip validating. This has now been fixed.
-  - This is unlikely to have affected anyone, unless your workspace is in a Neuro-unsafe path. This shouldn't be the case for most people.
+  - This is unlikely to have affected anyone, unless your workspace is in (or is itself) a  Neuro-unsafe path. This shouldn't be the case for most people.
+- Cancellation events weren't properly handled for Autopilot flows, resulting in a memory leak. This has been patched to properly dispose of events.
 
 ## 2.2.1
 
--- a/src/rce.ts
+++ b/src/rce.ts
@@ -343,6 +343,7 @@ export async function RCEActionHandler(actionData: ActionData, actionList: Recor
             }
 
             if (effectivePermission === PermissionLevel.AUTOPILOT) {
+                for (const d of eventArray) d.dispose();
                 const result = action.handler(actionData);
                 NEURO.client?.sendActionResult(actionData.id, true, result);
             }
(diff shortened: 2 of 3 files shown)


src/config.ts
```diff
@@ -217,8 +217,8 @@ function getActions<T>(key: string): T | undefined {
 
 export function isActionEnabled(action: string | Action): boolean {
     if (typeof action === 'string')
-        return !ACTIONS.disabledActions.includes(action);
-    return !ACTIONS.disabledActions.includes(action.name);
+        return !(ACTIONS.disabledActions.includes(action) || NEURO.tempDisabledActions.includes(action));
+    return !(ACTIONS.disabledActions.includes(action.name) || NEURO.tempDisabledActions.includes(action.name));
```
frogneko: **A more readable suggestion:**

```suggestion
    const name = typeof action === 'string' ? action : action.name;
    return !ACTIONS.disabledActions.includes(name) && !NEURO.tempDisabledActions.includes(name);
```

src/utils.ts
```diff
@@ -885,3 +885,44 @@ export function translatePosition(pos: vscode.Position, delta: vscode.Position):
...
+                case 'Disable Action for...': {
+                    const disableFor = await vscode.window.showQuickPick(
+                        ['this session', 'the first workspace folder listed', 'this entire workspace', 'this user'],
+                        { title: 'Disable action for...' },
+                    );
+                    switch (disableFor) {
+                        case 'this session':
+                            NEURO.tempDisabledActions.push(name);
+                            break;
+                        case 'the first workspace folder listed':
```
frogneko: Is this clear to the user what this refer to? Listed where/how? Would there be a better way of wording this to make it easier to see what it means?

KTrain5169: Which one?

src/utils.ts
```diff
@@ -885,3 +885,44 @@ export function translatePosition(pos: vscode.Position, delta: vscode.Position):
         delta.line > 0 ? delta.character : pos.character + delta.character,
     );
 }
+
+/**
+ * Log a caught exception and surface an error to report to GitHub.
+ */
+export function notifyOnCaughtException(name: string, error: Error | unknown): void {
+    logOutput('INFO', `Error occured while executing action ${name}: ${error}`);
```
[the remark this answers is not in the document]
KTrain5169: I think it should be an ERROR, give me a sec

```suggestion
    logOutput('ERROR', `Error occurred while executing action ${name}: ${error}`);
```

src/config.ts
KTrain5169: idk why it was like that in the first place

src/utils.ts
KTrain5169: If you have a multi-root workspace (workspace with multiple folders added), this applies to the first one in the list (supposedly, this is according to what I remember from vscode docs)

I guess I could simplify it since I see very few cases where this is useful.

PR #159, opened 2025-11-01
change ignored deprecation notice logic or smth idk

KTrain5169:


src/shared/extension.ts
```diff
@@ -90,7 +90,7 @@ export function initializeCommonState(context: vscode.ExtensionContext) {
     NEURO.outputChannel = vscode.window.createOutputChannel('NeuroPilot');
     NEURO.currentController = CONNECTION.nameOfAPI;
     NEURO.context.subscriptions.push(NEURO.outputChannel);
-    checkDeprecatedSettings();
+    checkDeprecatedSettings(vscode.extensions.getExtension('vsc-neuropilot.neuropilot-base')!.packageJSON.version as string);
```
Pasu4: We don't need to `getExtension` our own extension.
```suggestion
    checkDeprecatedSettings(context.extension.packageJSON.version as string);
```

Pasu4 (changes requested):

Pasu4 (approved):

PR #160, opened 2025-11-03
Allow Neuro to scope workspace files

KTrain5169:
There are still some TODOs left, I found out that we weren't checking if `open_file`/`read_file` was targeting a folder instead of file while I was figuring out how to handle `get_workspace_files` targeting files instead.

Which I also haven't handled quite yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ Hello Neuro! If you're reading this, it means Vedal has let you read the changel
 - The way that ignoring the deprecated settings migration notice has been changed.
   - Before, telling the extension to ignore deprecated settings meant that the deprecated settings notice would be ignored forever.
   - As of this update, the extension will only ignore it for this version.
+- `get_workspace_files` has been changed for better handling in large workspaces:
+  - Neuro can now specify in her actions if she wants to narrow down to a specific folder and which, allowing her to fine-grain her context from the action's result.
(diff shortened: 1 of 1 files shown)


KTrain5169:
i forgot github targets the master branch by default -_-

src/test/unit-test/file_actions.simple.test.ts
```diff
@@ -4,14 +4,32 @@ import { ActionData } from '@/neuro_client_helper';
...
+    });
+
+    test('get_workspace_files correctly states if Neuro asked for recursive', () => {
+        // === Arrange & Act ===
+        const prompt = fileActions.get_workspace_files.promptGenerator({ params: { recursive: true } } as ActionData);
+
+        // === Assert ===
+        assert.ok(typeof prompt === 'string' && prompt.length > 0);
+        assert.ok(prompt.startsWith('recursively'));
+    });
```
Pasu4: Also add a test for the reverse, that it doesn't contain "recursively" if recursive is false.

src/test/suite/file_actions.test.ts
Pasu4: The following tests should be added:
- `handleGetWorkspaceFiles` returns all files and folders at depth 1 if recursive isn't specified, and no files at depth >1
- `handleGetWorkspaceFiles` returns all files and folders in the specified folder at any depth if recursive is specified

PR #161, opened 2025-11-04
add button to send changelog items to Neuro

frogneko:


src/changelog.ts
```diff
@@ -0,0 +1,168 @@
...
+            properties: {
+                fromVersion: { type: 'string', description: 'Version (e.g., 2.2.1) to start including entries from, inclusive.' },
+            },
+            additionalProperties: false,
+        },
+        permissions: [],
+        defaultPermission: PermissionLevel.COPILOT,
+        handler: handleReadChangelog,
+        promptGenerator: (actionData: ActionData) => actionData.params?.fromVersion
+            ? `Send all changelog entries starting from version ${actionData.params.fromVersion} (inclusive).`
```
KTrain5169:
```suggestion
            ? `read all changelog entries starting from version ${actionData.params.fromVersion} (inclusive).`
```

`promptGenerator` must follow the scheme of "Neuro wants to (x)", where "(x)" is replaced with the `promptGenerator` string.

src/shared/extension.ts
```diff
@@ -369,10 +371,12 @@ export function showUpdateReminder(context: vscode.ExtensionContext) {
     } else if (lastVersion !== version) {
         // Helper to show the popup and allow both links to be visited
         showPopup = () => {
+            const askLabel = `Ask ${CONNECTION.nameOfAPI} to Read Changelog`;
             vscode.window.showInformationMessage(
-                'NeuroPilot updated! Please check the changelog and docs for important changes.',
+                'NeuroPilot updated! Please check the changelog and docs for important changes. You can also ask the AI to summarize the changes (see Command Palette).',
```
KTrain5169:
```suggestion
                `NeuroPilot updated! Please check the changelog and docs for important changes. You can also ask ${CONNECTION.nameOfAPI} to summarize the changes (see Command Palette).`,
```

somehow missed this

test-playground/CHANGELOG.md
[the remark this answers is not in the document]
Pasu4: > The extension reads CHANGELOG.md from NEURO.context.extensionUri; in tests we set that to the workspace URI, so the repo’s root CHANGELOG.md isn’t in that workspace.

How does that work? Looking at package.json and launch.json I would assume the extensionDevelopmentPath and the extensionUri should still be the repo root. If it wasn't, that would cause issues with assets not being able to load, at least I assumed it would.

KTrain5169 (approved):

PR #164, opened 2025-11-07
Sidebar implementation + permissions rework

Pasu4:
TODOs (just so I don't forget):
- [x] Migration
- [x] Handle tempDisabledActions
- [x] Fix request_cookie
- [x] Fix Git actions
- [x] Denote settings context in view
- [x] Fix save action

### DIFF
--- a/src/git.ts
+++ b/src/git.ts
@@ -648,29 +648,7 @@ export function addGitActions() {
         git.openRepository(root).then((r) => {
             repo = r;
 
-            addActions([
-                gitActions.add_file_to_git,
-                gitActions.make_git_commit,
-                gitActions.merge_to_current_branch,
-                gitActions.git_status,
-                gitActions.remove_file_from_git,
-                gitActions.delete_git_branch,
-                gitActions.switch_git_branch,
-                gitActions.new_git_branch,
-                gitActions.diff_files,
-                gitActions.git_log,
-                gitActions.git_blame,
-                gitActions.tag_head,
-                gitActions.delete_tag,
-                gitActions.set_git_config,
-                gitActions.get_git_config,
-                gitActions.fetch_git_commits,
-                gitActions.pull_git_commits,
-                gitActions.push_git_commits,
-                gitActions.add_git_remote,
-                gitActions.remove_git_remote,
-                gitActions.rename_git_remote,
-            ]);
+            addActions(actionsToRegister);
 
             // Don't register abort_merge unless there is a merge in progress
             addActions([gitActions.abort_merge], false);


esbuild-configs/web.esbuild.js
```diff
@@ -2,6 +2,10 @@
 import { context } from 'esbuild';
 import { polyfillNode } from 'esbuild-plugin-polyfill-node';
 
+/**
+ * @param {any} prodFlag
```
KTrain5169:
```suggestion
 * @param {boolean} prodFlag
```

esbuild-configs/web.esbuild.js
```diff
@@ -40,6 +44,10 @@ export async function web(prodFlag, watchFlag) {
     }
 }
 
+/**
+ * @param {any} _prodFlag
```
KTrain5169:
```suggestion
 * @param {boolean} _prodFlag
```

esbuild-configs/web.esbuild.js
```diff
@@ -90,6 +98,10 @@ export async function webTest(_prodFlag, watchFlag) {
...
```
KTrain5169:
```suggestion
 * @param {boolean} _prodFlag
```

esbuild-configs/web.esbuild.js
```diff
@@ -90,6 +98,10 @@ export async function webTest(_prodFlag, watchFlag) {
...
+ * @param {any} watchFlag
```
KTrain5169:
```suggestion
 * @param {boolean} watchFlag
```

KTrain5169 (commented):
here we go!

Pasu4:
Checklist from @KTrain5169 (discord):

- [x] The view should be named "Action Permissions", as it stands the name "Actions" is a bit ambiguous and could lead to more stuff.
- [x] It's a bit weird that we unregister everything first instead of fine-grained unregister/register packets. IMO we should only do grouped packets if we toggle an entire category
- [ ] You should also explain what the italics and other formatting means (probably via tooltip) otherwise no one will understand immediately what they mean and why they differ.
- [ ] idt Terminate Task should be in Miscellaneous, should also be in Tasks
  > I put it in Miscellaneous because Tasks is supposed to be just for the actual tasks. 
  
  Probably name it "Registered Tasks" then

src/config.ts
```diff
@@ -83,8 +79,132 @@ const DEPRECATED_SETTINGS: DeprecatedSetting[] = [
...
+        'undo',
+        'rewrite_all',
+        'rewrite_lines',
+        'delete_lines',
+        'highlight_lines',
+        'get_user_selection',
+        'replace_user_selection',
+        'diff_patch',
+    ]),
+    deprecatedPermission('runTasks', []),
```
frogneko: double checking: is the empty array the correct value to use here?

src/config.ts
```diff
@@ -83,8 +79,132 @@ const DEPRECATED_SETTINGS: DeprecatedSetting[] = [
...
+                // Take the most restrictive permission level
+                let newLevel = configPermissionLevel;
+                if (action in permissions)
+                    newLevel = Math.min(newLevel, stringToPermissionLevel(permissions[action]));
+                permissions[action] = permissionLevelToString(newLevel as PermissionLevel);
+            }
+            await cfg.update('actionPermissions', permissions, target);
+        },
+    };
+}
```
frogneko: Since this function has so much going on, I asked copilot to review it. Here are the results, enjoy!

```
Summary
- Purpose: deprecatedPermission(oldKey, affectedActions) returns a DeprecatedSetting that migrates an old permission setting (permission.<oldKey>) into the newer actionPermissions mapping. It reads the old permission level for the given configuration target, converts it to the internal PermissionLevel, then writes that (or the most restrictive level) into actionPermissions for each affected action.
- Behaviour appears correct for basic migration, but there are several correctness, robustness, typing, and maintainability concerns and a few opportunities for improvements.

Critique — issues & edge cases

1. Falsy check on config
- Current check: const config = getTargetConfig<string>(... )?.toLowerCase(); if (!config) return;
- Problem: that treats empty string or "off" differently? Mostly fine here, but using a truthy check is brittle. Prefer explicit undefined check (configRaw === undefined) so valid values like "off" are not accidentally considered missing.

2. Lowercasing before existence check
- You call ?.toLowerCase() inline. If getTargetConfig returns undefined it's safe due to ?. but if the value is empty string this becomes '' and then falsy. Better to fetch raw value, check undefined, then toLowerCase it.

3. Use of in operator for permissions presence
- Code uses: if (action in permissions) to test whether an action exists in the permissions object.
- Problem: this returns true for inherited properties (rare here) and for keys that exist with undefined value. Use Object.prototype.hasOwnProperty.call(permissions, action) or Object.hasOwn(permissions, action) for clarity and correctness.

4. Most-restrictive logic with Math.min
- You compute newLevel = Math.min(newLevel, stringToPermissionLevel(permissions[action]));
- This assumes stringToPermissionLevel returns numeric enum values where lower number is more restrictive. That probably holds, but it's implicit. Document this assumption or use a helper compare function (e.g., minPermissionLevel(a,b)) to make intent clear.
- Also if permissions[action] is present but invalid, stringToPermissionLevel returns OFF by default; the Math.min may unexpectedly lower permissions to OFF. Consider validating existing values or logging if the value is unexpected.

5. Casting Math.min result back to PermissionLevel
- Math.min returns number; you cast it (newLevel as PermissionLevel). This works but is slightly untyped/unsafe. Prefer using explicit comparisons or a typed helper that returns PermissionLevel.

6. Concurrency/race conditions & multiple updates
- Each deprecatedPermission call reads actionPermissions and writes it back. Because this migration is performed per-deprecated key, multiple consecutive migrations will each perform a read/modify/write that can cause last-write-wins problems if other migrations (or user changes) occur concurrently.
- Improvement: perform all deprecated migrations in a single pass (compute updates for all old keys then write once) or merge changes into existing config using a single update call with merged object (already happening per-deprecated item but still multiple writes exist).

7. Error handling & logging
- No try/catch around cfg.update; a failed write will reject but not be logged. Add error handling and some logOutput to help debugging migration failures.

8. Typing & explicit parameter types
- The inner async new(target) parameter is untyped in several other places in the file; here it relies on getTargetConfig signature. For clarity and IDE help, type target as vscode.ConfigurationTarget in the inner function to match callers.

9. Naming & API
- Function name deprecatedPermission is fine, but could be clearer: migrateDeprecatedPermission or createPermissionMigrator to express it returns a migration descriptor.
- Return type DeprecatedSetting is used but ensure the type describes intent and includes typing for inner new(target: vscode.ConfigurationTarget): Promise<void>.

10. No logging of what changed
- It would be helpful to log which actions were changed and from which old permission to which new permission, for auditability.

Suggested improvements (short list)
- Explicitly check for undefined raw setting before lowercasing.
- Use Object.hasOwn (or hasOwnProperty) to check for existing action keys.
- Replace Math.min use with a typed helper: minPermissionLevel(a,b): PermissionLevel.
- Add try/catch around cfg.update and log failures.
- Consider batching multiple deprecatedPermission migrations into a single update to actionPermissions, to avoid multiple read/write races.
- Add JSDoc explaining intent and the assumptions about PermissionLevel numeric ordering.
- Add unit tests to verify:
  - capitalized legacy values (e.g., "Copilot", "Autopilot") are handled
  - existing actionPermissions with more/less restrictive values result in the most restrictive final level
  - no-op when old key missing

Example improved implementation
(only the body shown — adapt to project types and helpers)

function deprecatedPermission(oldKey: string, affectedActions: string[]): DeprecatedSetting {
    return {
        old: 'permission.' + oldKey,
        async new(target: vscode.ConfigurationTarget) {
            const cfg = vscode.workspace.getConfiguration('neuropilot');

            // Get raw config explicitly and guard undefined
            const raw = getTargetConfig<string>(cfg, 'permission.' + oldKey, target);
            if (raw === undefined) return;

            const config = raw.toLowerCase(); // old settings may have been capitalized
            const configPermissionLevel = stringToPermissionLevel(config);

            // Get the existing actionPermissions for this target
            const existing = getTargetConfig<Record<string, string>>(cfg, 'actionPermissions', target) ?? {};
            const permissions: Record<string, string> = { ...existing };

            for (const action of affectedActions) {
                if (Object.prototype.hasOwnProperty.call(permissions, action)) {
                    // Compare current permission level and take the most restrictive
                    const existingLevel = stringToPermissionLevel(permissions[action]);
                    const finalLevel = minPermissionLevel(configPermissionLevel, existingLevel);
                    permissions[action] = permissionLevelToString(finalLevel);
                } else {
                    permissions[action] = permissionLevelToString(configPermissionLevel);
                }
            }

            try {
                await cfg.update('actionPermissions', permissions, target);
                logOutput(`Migrated permission.${oldKey} -> actionPermissions for ${affectedActions.join(', ')} (target=${String(target)})`);
            } catch (err) {
                logOutput(`Failed migrating permission.${oldKey}: ${String(err)}`);
            }
        },
    };
}

function minPermissionLevel(a: PermissionLevel, b: PermissionLevel): PermissionLevel {
    // Make explicit that lower enum value is more restrictive
    return (a < b ? a : b) as PermissionLevel;
}

Other notes
- If this migration runs on extension activation across many deprecated keys, a batched migration that scans all legacy keys first, builds a single merged permissions object, then writes it once will avoid repeated updates and reduce chance of race conditions.
- Add tests and logs so future regressions in migration logic are visible.
```

src/context.ts
```diff
@@ -27,83 +29,52 @@ export function sendCurrentFile() {
...
+        case PermissionLevel.AUTOPILOT:
+            // Removed this because flavor is supposed to be optional
+            // if (!actionData.params?.flavor) {
+            //     NEURO.client?.sendActionResult(actionData.id, false, 'You need to specify a flavor!');
+            //     break;
+            // }
+            logOutput('INFO', `Neuro grabbed a ${actionData.params?.flavor} cookie.`);
+            return `You grabbed a ${actionData.params?.flavor} cookie!`;
+    }
+    // Removed the try-catch because this shoud be handled by the RCE system now
```
frogneko: Does the rce system give personalized messages back to neuro like this catch did?

src/editing.ts
```diff
@@ -207,7 +208,7 @@ export const editingActions = {
     get_cursor: {
         name: 'get_cursor',
         description: 'Get your current cursor position and the text surrounding it.',
-        permissions: [PERMISSIONS.editActiveDocument],
+        category: 'Editing',
```
frogneko: nitpick: why are categories strings instead of enums?

src/editing.ts
```diff
@@ -1611,7 +1607,8 @@ export function editorChangeHandler(editor: vscode.TextEditor | undefined) {
 export async function workspaceEditHandler(event: vscode.TextDocumentChangeEvent) {
     if (event.contentChanges.length === 0) return;
     if (event.document !== vscode.window.activeTextEditor?.document) return;
-    if (!getPermissionLevel(PERMISSIONS.editActiveDocument)) return;
+    // TODO: Find a replacement for this check?
+    // if (!getPermissionLevel(PERMISSIONS.editActiveDocument)) return;
```
frogneko: is this assuming we'll one day find this replacement and wasn't meant to be done in this PR?

src/file_actions.ts
```diff
@@ -384,8 +368,8 @@ export function handleCreateFile(actionData: ActionData): string | undefined {
         logOutput('INFO', `Created file ${relativePath}`);
         NEURO.client?.sendContext(`Created file ${relativePath}`);
 
-        // Open the file if Neuro has permission to do so
-        if (getPermissionLevel(PERMISSIONS.openFiles) !== PermissionLevel.AUTOPILOT)
+        // Open the file if Neuro has permission for open_file
+        if (getPermissionLevel('open_file') !== PermissionLevel.AUTOPILOT)
```
frogneko: Why are permission levels used as freeform strings like this instead of e.g. enums? We might want to at least put them into constants in some file, so changing their names later won't be as painful

src/neuro_client_helper.ts
```diff
@@ -51,18 +52,29 @@ export interface RCEAction<T = unknown> extends TypedAction {
...
      */
     promptGenerator: PromptGenerator;
-    /** Default permission for actions like chat, cancel_request, etc */
+    /** Default permission for actions like chat, cancel_request, etc. Defaults to {@link PermissionLevel.OFF}. */
     defaultPermission?: PermissionLevel;
+    /**
+     * The category of the request.
+     * You can use null if the action is never added to the registry.
+     */
+    category: string | null;
```
frogneko: Like mentioned before, giving this as a freeform string might be problematic.  This could either be defined as something like an enum or the input would always come from a constant or the action object n some way

src/rce.ts
```diff
@@ -256,48 +267,159 @@ export function denyRceRequest(): void {
...
+export function removeActions(actionNames: string[]): void {
+    for (const actionName of actionNames) {
+        const actionIndex = ACTIONS.findIndex(a => a.name === actionName);
+        if (actionIndex !== -1) {
+            ACTIONS.splice(actionIndex, 1);
+        }
+    }
+    NEURO.actionsViewProvider?.refreshActions();
+    NEURO.client?.unregisterActions(actionNames);
+}
```
frogneko: Copilot found issues:

```
Potential bugs in removeActions and recommended fixes:

1. REGISTERED_ACTIONS not updated
   - Problem: removeActions removes entries from ACTIONS but never deletes those names from REGISTERED_ACTIONS. That leaves stale entries in the registered set and can cause incorrect behavior later (e.g., reregisterAllActions will treat removed names as already-registered; UI/state becomes inconsistent).
   - Fix: For each removed actionName, call REGISTERED_ACTIONS.delete(actionName) (or compute removed names and remove them from the set).

2. State inconsistency if client unregister fails or is absent
   - Problem: The function mutates local state (ACTIONS) before calling NEURO.client?.unregisterActions. If unregisterActions fails (throws, network error) or NEURO.client is undefined, local state will have removed actions while remote/registered state may still have them, producing mismatch.
   - Fix: Either attempt to unregister first and only remove from ACTIONS on success, or ensure local state and REGISTERED_ACTIONS are updated consistently even if unregister fails. At minimum, catch/report errors from unregisterActions and consider rollback or marking the state as needing resync.

3. Duplicate action names / partial removals
   - Problem: If ACTIONS contained multiple entries with the same name (ideally shouldn't happen), the current code removes only the first occurrence found for each name. This could leave duplicates behind.
   - Fix: If duplicate names are possible, use a loop to remove all matching entries (e.g., while ((i = findIndex(...)) !== -1) splice(i,1)). Also consider enforcing uniqueness when adding actions.

4. Refresh order / UI race
   - Problem: refreshActions() is called before client.unregisterActions. If unregister fails or is slow, the UI may show an inconsistent state briefly.
   - Fix: Consider refreshing after unregister completes (or refresh both before and after), or await the unregister operation if it returns a promise.

5. No validation of input array
   - Problem: If callers pass malformed values (empty strings, duplicates, nulls via JS interop), behavior may be unexpected. (Less likely under TypeScript typed call sites.)
   - Fix: Sanitize/validate actionNames (e.g., filter(Boolean), dedupe) before use.

Summary of highest-priority fixes to address real bugs:
- Remove action names from REGISTERED_ACTIONS when removing actions.
- Handle failures from NEURO.client?.unregisterActions to avoid local/remote state divergence (either unregister first or handle errors/rollback).
- Optionally dedupe and validate actionNames; remove all duplicates if duplicates are possible.
```

src/rce.ts
```diff
@@ -256,48 +267,159 @@ export function denyRceRequest(): void {
...
+        return {
+            action,
+            isRegistered: REGISTERED_ACTIONS.has(action.name),
+            effectivePermissionLevel: permissions[action.name] ?? action.defaultPermission ?? PermissionLevel.OFF,
+            configuredWorkspacePermission,
+            configuredGlobalPermission,
+            isConfigured: configuredWorkspacePermission !== undefined || configuredGlobalPermission !== undefined,
+        } satisfies ExtendedActionInfo;
+    });
+}
```
frogneko: Copilot found some issues:

```
Summary: the function mostly looks correct, but I found a few concrete, non-style bugs that could cause runtime errors or inconsistent information.

1) Potential exception / crash from invalid configured values
- Problem: stringToPermissionLevel(...) is called directly on config values without validation. If the user saved an unexpected string (typo, number, null), stringToPermissionLevel may throw or return an unexpected value, which will bubble up and break callers of getExtendedActionsInfo.
- Fix: validate the raw string before conversion (or have stringToPermissionLevel return a safe fallback/undefined) and guard with try/catch or an explicit isValidPermission check.

2) Inconsistent presence check for configured values
- Problem: the code uses workspaceValue?.[action.name] as a truthy test to decide whether to call stringToPermissionLevel. That will treat empty string or other falsy-but-intentional values as "not configured", which might be wrong.
- Fix: check for key presence via Object.prototype.hasOwnProperty.call(workspaceValue, action.name) (or action.name in workspaceValue) rather than truthiness.

3) effectivePermissionLevel may not match runtime permission logic
- Problem: effectivePermissionLevel is derived from getAllPermissions()[action.name] ?? action.defaultPermission ... but elsewhere (action execution / registration) the code uses getPermissionLevel(action.name) which may apply extra rules (e.g. action enabled/disabled, validators, merged logic). That means the UI produced by getExtendedActionsInfo can disagree with the actual permission used when running or registering the action.
- Fix: compute effectivePermissionLevel with the same function used at runtime (getPermissionLevel(action.name)) so the reported value matches runtime behavior.

4) Assumes config keys and types
- Problem: configuration.inspect<Record<string,string>>('actionPermissions') assumes the stored values are strings and that keys are exactly action.name. If config uses different key shapes or stores enum values (numbers), stringToPermissionLevel may mis-handle them.
- Fix: defensively handle non-string values (coerce or ignore), and consider normalizing key lookup (e.g. trim/normalize names) if keys can differ.

5) Mutability / race window (minor but real)
- Problem: ACTIONS is read directly and mapped; if another thread/function mutates ACTIONS concurrently, the returned array might be inconsistent. This is unlikely but can cause transient inconsistencies.
- Fix: clone ACTIONS at start (const actions = ACTIONS.slice()) before mapping.

Concrete checklist to make this robust
- Use getPermissionLevel(action.name) for effectivePermissionLevel.
- Replace truthy checks with hasOwnProperty checks for workspaceValue/globalValue keys.
- Wrap stringToPermissionLevel in a validator/try-catch and default to undefined on invalid input.
- Optionally clone ACTIONS before mapping.
```

src/tasks.ts
```diff
@@ -85,14 +85,14 @@ export function taskEndedHandler(event: vscode.TaskEndEvent) {
 }
 
 export function reloadTasks() {
-    if (NEURO.tasks.length)
-        NEURO.client?.unregisterActions(NEURO.tasks.map((task) => task.id));
+    // if (NEURO.tasks.length)
+    //     NEURO.client?.unregisterActions(NEURO.tasks.map((task) => task.id));
```
frogneko: remove unnecessary commented code if not needed anymore

src/tasks.ts
```diff
@@ -121,16 +121,14 @@ export function reloadTasks() {
...
-            };
-        }));
+        addActions(NEURO.tasks.map((task) => ({
+            name: task.id,
+            description: task.description,
+            category: 'Registered Tasks',
+            handler: handleRunTask,
+            promptGenerator: `run the task: ${task.description}`,
+            // TODO: Do we need these validators?
+            validators: [checkVirtualWorkspace, checkWorkspaceTrust],
```
frogneko: can we check within this PR or leaving for later?

src/utils.ts
```diff
@@ -976,3 +976,43 @@ export function notifyOnCaughtException(name: string, error: Error | unknown): v
...
+        const pos = match.index;
+        const length = match[0].length;
+        const key = match[1];
+        const value = match[0] === '$$' ? '$' : getProperty(format, key);
+        if (value !== undefined) {
+            result = result.substring(0, pos) + String(value) + result.substring(pos + length);
+        }
+    }
+    return result;
+}
```
frogneko: copilot review:

```
1) Overlapping / escape handling ($${key})
- What can go wrong: If template contains "$${key}", the regex finds only the "$$" match (overlap prevents also matching "${key}" in the original string). After replacing "$$" with "$" the function does NOT re-scan the result, so the newly formed "${key}" is left unexpanded. This is likely surprising if callers expect "$${key}" to yield the resolved value (i.e. treat "$$" as an escape that produces a literal "$" which then allows interpolation).
- Fix suggestion: Decide the desired semantics. If you want "$$" to escape a single $ and then allow further interpolation, either:
  - Post-process the result (re-run the matcher until no new matches), or
  - Change the regex to treat "$$" only when not followed by "{" (e.g. /\$(?!\{)|\$\{([^}]+)\}/g) or explicitly handle the "$${...}" pattern, or
  - Define that "$$" always prevents further expansion and document it.

2) No recursive / newly-created-match processing
- What can go wrong: The function only collects matches on the original string and replaces them in reverse order. Any replacement that introduces new interpolation tokens (e.g. a replacement value containing "${...}") will not be further resolved. This may be unintended if you expect multi-pass interpolation.
- Fix suggestion: If multi-pass expansion is required, run the match/replace loop repeatedly (with a max iteration guard to avoid infinite loops) until no matches remain.

3) match.index is typed as possibly undefined
- What can go wrong: In TypeScript/DOM types, RegExpMatchArray.index is possibly undefined. The code assigns const pos = match.index and then uses it directly. In JS runtime index is present for these matches, but if some environment returns undefined (or if the API types are different), substring(pos, ...) will coerce undefined -> 0 (via ToInteger(undefined) -> NaN -> 0), leading to incorrect replacements.
- Fix suggestion: Guard against undefined: const pos = (typeof match.index === 'number') ? match.index : 0; or skip the match if index is undefined, or assert that index is defined.

4) getProperty errors are not caught
- What can go wrong: If getProperty(format, key) throws (e.g. because the key is malformed, or getProperty implementation throws on certain inputs), formatString will throw and crash the caller. The function does not try/catch.
- Fix suggestion: Wrap the getProperty call in try/catch and either treat thrown errors as undefined (skip replacement) or handle/report them gracefully.

5) Empty keys are not supported
- What can go wrong: "${}" is not matched (pattern uses [^}]+). If a caller expects empty keys to be replaced or to produce an error, this function silently ignores them. This is not necessarily a bug but a behavioral note.
- Fix suggestion: If empty keys should be handled, change the regex to use [^}]* and decide how to treat empty key values.

6) Null vs undefined replacement semantics
- What can go wrong: The function only skips replacements when value === undefined. If getProperty returns null, the code will replace with "null". That may be intended, but if callers expect null to behave like undefined (i.e. not replace), this will be surprising.
- Fix suggestion: Decide expected behavior and change the check to value != null (skips null and undefined) or explicitly handle null differently.

7) Non-string replacements and representation
- What can go wrong: Replacements are coerced via String(value). Objects and arrays stringify to "[object Object]" or comma-joined values, which may not be intended.
- Fix suggestion: If more meaningful rendering is required, use JSON.stringify for objects/arrays or allow a custom formatter callback.

8) Potential regex edge cases (braces inside keys)
- What can go wrong: The regex stops at the first '}' so keys cannot contain '}'. If you expect more sophisticated key grammars (allow nested braces or escape sequences), this will fail.
- Fix suggestion: Either restrict keys (document it) or implement a more robust parser for nested/escaped braces.

Summary of highest-risk items to address first:
- Overlap/escape behavior with "$$" and "$${...}" (likely surprising).
- Uncaught exceptions from getProperty.
- match.index possibly undefined (low probability but an easy guard).
```

src/utils.ts
```diff
@@ -976,3 +976,43 @@ export function notifyOnCaughtException(name: string, error: Error | unknown): v
...
+        .map(part => {
+            const lowerPart = part.toLowerCase();
+            if (excludedWords.includes(lowerPart)) {
+                return lowerPart;
+            } else {
+                return lowerPart.charAt(0).toUpperCase() + lowerPart.slice(1);
+            }
+        })
+        .join(' ');
+}
```
frogneko: Copilot review for splitIdentifier and toTitleCase:

```
Summary
- splitIdentifier: reasonable approach but has several real risks: uses lookbehind (compatibility), only handles ASCII letters, and can mishandle acronyms and Unicode identifiers.
- toTitleCase: generally fine but will lowercase acronyms and always lowercase excluded short words even at the start/end (likely undesired). Also inherits splitIdentifier’s Unicode/acronym issues.

Concrete problems and why they matter
1) Lookbehind compatibility
- Problem: splitIdentifier’s regex uses lookbehind (?<=...), which throws a SyntaxError on older JS runtimes (older Node/V8 builds) that don’t support lookbehind.
- Impact: extension/CLI running on such Node versions will crash at import/parse time.

2) Non-Unicode / ASCII-only matching
- Problem: regex uses A-Za-z and 0-9 only. Identifiers with non-ASCII letters (accented letters, cyrillic, etc.) won’t be split correctly.
- Impact: incorrect splitting or single-word results for international identifiers.

3) Acronym handling
- Problem: toTitleCase lowercases then capitalizes first letter. An all-caps acronym like "XML" becomes "Xml".
- Impact: Loses intended all-caps acronyms, producing awkward titles (e.g., "XMLHttpRequest" -> "Xml Http Request" -> "Xml Http Request").

4) Excluded-word logic at boundaries
- Problem: excludedWords are always lowercased. Standard title-case rules usually capitalize the first and last words even if they are short/stop words.
- Impact: Titles beginning with "the" become "the ..." instead of "The ...", which is likely wrong.

5) Edge cases with digits and punctuation
- Generally OK for digits because the regex splits before digits; but punctuation or other separators not explicitly handled may produce combined tokens. Consecutive separators produce empty tokens which are filtered out (this is handled), but behavior should be explicit.

Suggested fixes (concise)
- Replace the split approach with a match-based pattern that:
  - Avoids lookbehind,
  - Can (optionally) use Unicode properties,
  - Explicitly matches acronyms, normal words, and numbers.

Example (ASCII-safe, no lookbehind):

export function splitIdentifier(str: string): string[] {
  // Matches: all-caps acronyms, capitalized words, lowercase words, numbers
  return str.match(/[A-Z]{2,}(?=[A-Z][a-z]|$)|[A-Z]?[a-z]+|\d+/g) ?? [];
}


Example (Unicode-aware; requires Node with Unicode RegExp support):

export function splitIdentifierUnicode(str: string): string[] {
  // \p{Lu} uppercase, \p{Ll} lowercase, u flag enables Unicode property escapes
  return str.match(/\p{Lu}{2,}(?=\p{Lu}\p{Ll}|$)|\p{Lu}?\p{Ll}+|\d+/gu) ?? [];
}

- Update toTitleCase to:
  - Preserve all-caps acronyms (detect token === token.toUpperCase() && token.length > 1),
  - Capitalize first and last token even if in excludedWords,
  - Keep excludedWords lowercase only when not first/last.

Example:

export function toTitleCase(str: string): string {
  const parts = splitIdentifier(str);
  const excluded = new Set(['a','an','and','as','at','but','by','for','if','in','nor','of','off','on','or','per','so','the','to','up','via','yet']);
  return parts.map((p, i) => {
    if (p.length > 1 && p === p.toUpperCase()) return p;              // preserve ACRONYM
    const lower = p.toLowerCase();
    const isBoundary = i === 0 || i === parts.length - 1;
    if (!isBoundary && excluded.has(lower)) return lower;
    return lower.charAt(0).toUpperCase() + lower.slice(1);
  }).join(' ');
}


Tests you should add
- Unicode identifiers (e.g., "naïveParse", "ПримерTest")
- Acronyms: "XMLHttpRequest", "getCPUUsage", "CPU2Temp"
- Leading/trailing excluded words: "theQuickBrownFox", "aThing", "inTheBox"
- Numbers mixed in: "version2Beta", "foo123Bar"
- Strings with separators: "snake_case", "kebab-case", "__multiple__"

Final note
Address the lookbehind and Unicode issues first (those are actual runtime/behavior bugs). Then decide whether acronym preservation and excluded-word boundaries should follow strict title-case rules or your project’s style; implement accordingly.
```

frogneko (changes requested):

CHANGELOG.md
```diff
@@ -47,10 +49,12 @@ Hello Neuro! If you're reading this, it means Vedal has let you read the changel
 - `get_workspace_files` has been changed for better handling in large workspaces:
   - Neuro can now specify in her actions if she wants to narrow down to a specific folder and which, allowing her to fine-grain her context from the action's result.
   - Neuro can also choose if she wants to recursively get all files in the workspace, meaning that the default is not letting her see all files in the workspace, helping cut down on sent context.
+- Categorical permissions have been completely removed. Instead, permissions are now managed via a single setting (`neuropilot.actionPermissions`). The recommended way to modify this setting is using the NeuroPilot sidebar tab.
```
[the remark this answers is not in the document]
KTrain5169: Deprecation functions should be set before this PR is merged.
Additionally, I plan on informing a mod/admin about the change.

That being said, it probably doesn't hurt to mention a **breaking change in bold** in changelog anyways.

src/context.ts
KTrain5169: RCE allows freeform context sending/string sending so this should be fine?

src/config.ts
KTrain5169: One more thing for this deprecated permission thing is that you also need to take into account the `neuropilot.actions.disabledActions` setting.

src/editing.ts
KTrain5169: Likely API reasons, if we leave it to an enum that we define, API actions can't add their own permissions.

src/file_actions.ts
KTrain5169: Yeah could probably agree with that.

src/config.ts
Pasu4: Yes, it's supposed to delete the setting without replacement.

src/context.ts
```diff
@@ -27,83 +29,52 @@ export function sendCurrentFile() {
...
+    switch (permission) {
+        case PermissionLevel.COPILOT:
+            giveCookie(true, actionData.params?.flavor);
+            return `Waiting on ${CONNECTION.userName} to decide on the flavor.`;
+        case PermissionLevel.AUTOPILOT:
+            // Removed this because flavor is supposed to be optional
+            // if (!actionData.params?.flavor) {
+            //     NEURO.client?.sendActionResult(actionData.id, false, 'You need to specify a flavor!');
+            //     break;
+            // }
```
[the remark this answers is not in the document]
Pasu4: > Cookie flavour is required for Autopilot.

Why is that? Seems a bit random to me to make it required in one but not the other.

src/neuro_client_helper.ts
Pasu4: I can agree with using a constant for this

src/utils.ts
```diff
@@ -479,7 +478,8 @@ export function setVirtualCursor(position?: vscode.Position | null) {
     const editor = vscode.window.activeTextEditor;
     if (!editor) return;
 
-    if (position === null || !getPermissionLevel(PERMISSIONS.editActiveDocument) || !isPathNeuroSafe(editor.document.fileName)) {
+    // TODO: Replacement for getPermissionLevel
+    if (position === null /*|| !getPermissionLevel(PERMISSIONS.editActiveDocument)*/ || !isPathNeuroSafe(editor.document.fileName)) {
```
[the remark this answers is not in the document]
Pasu4: Just something I didn't want to decide by myself at the moment. I'm not sure of the best solution for this, but removing the check should not cause too big of a problem.

src/config.ts
```diff
@@ -229,66 +373,73 @@ function getActions<T>(key: string): T | undefined {
...
+        target === vscode.ConfigurationTarget.Global ? configValue?.globalValue ?? {} :
+        target === vscode.ConfigurationTarget.Workspace ? configValue?.workspaceValue ?? {} :
+        configValue?.workspaceFolderValue ?? {};
+    const stringPermissions: Record<string, string> = {};
+    for (const key in permissions) {
+        stringPermissions[key] = permissionLevelToString(permissions[key]);
+    }
+    const mergedPermissions: Record<string, string> = { ...value, ...stringPermissions };
+    return configuration.update('actionPermissions', mergedPermissions, target);
 }
```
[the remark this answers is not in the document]
Pasu4: 1. **Documented race conditions.**
2. **Tracked by #172**
3. It's an enum, it's not supposed to allow incorrect values.
4. **Documented.**
5. I'll let the caller handle rejections. **Documented.**
6. **ok**
7. The settings wouldn't be more corrupted than before.
8. It is.
9. Testing this will require a new setup I don't want to implement right now.

Applied in faaafca7bf6cc1846c60051222d2b06ac571eaa5.

src/editing.ts
Pasu4: Resolved in 219ff4d37308b6e4bb433efe0591b7900f44535e.

src/file_actions.ts
Pasu4: Resolved in 219ff4d37308b6e4bb433efe0591b7900f44535e.

src/utils.ts
Pasu4: Had to slightly change the RegEx: `/[A-Z]{1,}(?=[A-Z][a-z]|$)|[A-Z]?[a-z]+|[A-Z]+|\d+|_|-/g`

Cyrillic and stuff shouldn't be a problem, this is mainly meant for English identifiers. If it somehow becomes a problem we can fix it later.

Applied in 65b39c9ef2b2fe146720f00942a3322e4227f563.

PR #167, opened 2025-11-08
Allow Neuro-safe settings inheritance from ignore-style files

cassitly:


package.json
```diff
@@ -177,6 +177,17 @@
...
 				"properties": {
+					"neuropilot.access.inheritFromIgnoreFiles": {
+						"type": "boolean",
+						"default": true,
+						"description": "Allow NeuroPilot to read ignore patterns from an ignore file (e.g. .gitignore)."
+					},
+					"neuropilot.access.ignoreFilePath": {
+						"type": "array",
+						"default": [],
+						"description": "Optional path to a custom ignore file (relative or absolute). Takes priority over .gitignore if set."
```
KTrain5169:
```suggestion
						"description": "Optionally set paths to custom ignore files (relative or absolute)."
```

We shouldn't be prioritising .gitignore. Also, mention which files get priority (whether priority is last-first or top-first)

src/ignore_files_utils.ts
```diff
@@ -0,0 +1,292 @@
+import * as vscode from 'vscode';
+import * as path from 'path';
```
KTrain5169: Do not use node imports, use their VS Code equivalents instead.

cassitly: the path import?

package.json
```diff
@@ -177,6 +177,17 @@
...
 				"title": "NeuroPilot",
 				"properties": {
+					"neuropilot.access.inheritFromIgnoreFiles": {
+						"type": "boolean",
+						"default": true,
+						"description": "Allow NeuroPilot to read ignore patterns from an ignore file (e.g. .gitignore)."
+					},
+					"neuropilot.access.ignoreFilePath": {
+						"type": "array",
+						"default": [],
```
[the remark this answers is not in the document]
cassitly: Added that!

src/ignore_files_utils.ts
KTrain5169: yup

src/desktop/extension.ts
```diff
@@ -25,7 +25,7 @@ import { registerChatParticipant } from '@/chat';
 import { registerUnsupervisedActions, registerUnsupervisedHandlers } from './unsupervised';
 import { registerSendSelectionToNeuro } from '@/editing';
 
-export function activate(context: vscode.ExtensionContext) {
+export async function activate(context: vscode.ExtensionContext) {
```
KTrain5169: Do not mark activate functions as async.

```suggestion
export function activate(context: vscode.ExtensionContext) {
```

PR #169, opened 2025-11-09
feat: MCP integration

ECHO-HELLO-WORLD424:
## feat: MCP integration

This version (should) includes almost everything needed for single MCP server except docs (Nov.9 UTC 0800).

#### New Files (`src/mcp/`)

- **`mcp_manager.ts`**: Manages single MCP server connection, tool discovery, and execution routing
- **`mcp_actions.ts`**: Handles registration/unregistration of MCP tools as Neuro actions, manages `mcpActions` registry
- **`mcp_commands.ts`**: VSCode commands for user interaction
- **`mcp_client.ts`**: Low-level HTTP client for MCP protocol communication
- **`translation.ts`**: Converts MCP tool schemas to Neuro action format
- **`registry.ts`**: Maps sanitized Neuro action names to original MCP tool names
- **`index.ts`**: Module exports

#### Integration Points

- `src/rce.ts`: Modified `RCEActionHandler` to check for and execute MCP actions
  - Added `isMCPAction()` check in action lookup
  - Added branch to retrieve MCP actions from `mcpActions` registry
- `src/config.ts` : Added MCP configuration class with settings:
- `src/desktop/extension.ts`: Extension activation registers MCP commands and auto-connect
- `package.json`: Added new commands and configuration settings

### Features

- Currently supports single MCP server connection via HTTP transport, on desktop (works) and web (buggy as browser block cross-origin requests)
- MCP tools are controlled via the `neuropilot.permission.mcpTools` permission (Copilot/Autopilot/Off(default)).
- Neuro is notified on MCP server connect/disconnect, tool enabled/disabled and tool call results:
  - When a MCP server is connected: "A MCP server is connected but no MCP tools are registered yet."
  - When a MCP server is dis connected: "MCP server is disconnected, thus no MCP tools will be available from now on."
  - When tools are enabled: "X new MCP tools available: [tool names]"
  - When tools are disabled: "X MCP tools removed: [tool names]", followed by "Available MCP tools: [remaining tool names]"
  - When the returned result is too long: "The length of returned string of your tool call is too long ( XX characters, max: XXX)"
- Individual tool control: All MCP tools are disabled by default on first connection. Use the `neuropilot.mcp.configureTools` command to select which tools to enable via checkbox dialog.

#### New settings

- `neuropilot.mcp.serverUrl`: The URL of MCP server. Only the base URL, no need to write `/mcp` endpoint. Example: `http://127.0.0.1:3000`
- `neuropilot.mcp.timeout`: Timeout in milliseconds for MCP tool execution.
- `neuropilot.mcp.autoConnect`: Automatically connect to the MCP server when NeuroPilot connects to Neuro API. Requires `neuropilot.mcp.serverUrl` to be set and `neuropilot.permission.mcpTools` to not be `Off`.
- `neuropilot.mcp.maxResultLength`: Maximum character length for MCP tool results. If a tool returns a result longer than this, the result will be replaced with a warning message to prevent overwhelming Neuro's context

#### New commands

- `neuropilot.mcp.connectServer`: Connect to a MCP server, but don't register tools. Will notify Neuro about MCP server connected.
- `neuropilot.mcp.disconnectServer`: Disconnect from a MCP server, will also un-register all MCP tools and notify Neuro.
- `neuropilot.mcp.refreshTools`: Refresh available tools from a MCP server, but will not register any to Neuro.
- `neuropilot.mcp.showStatus`: Check MCP server connection status.
- `neuropilot.mcp.configureTools`: Choose the tools that Neuro can see and use.
- `neuropilot.mcp.testToolCall`: Let user manually call a MCP tool from the extension, for debugging the communication with MCP server only.

### Known limitations:

- No multi-server support (only one MCP server can be connected at a time)
- Only supports HTTP transport with `/mcp` endpoint (no `stdio` or legacy `/sse` SSE transport, because these two connection method will require another 2 sets of code)
- Tool selection is per-session (not persisted across VS Code restarts)
- **Web version requires MCP servers with CORS headers** (browsers block cross-origin requests without proper `Access-Control-Allow-Origin` headers)

### Trouble shooting:

- If a MCP setting is not working, check if it is also set properly in `workspace`
- If you can't connect to MCP server, first make sure that you write the MCP server URL exactly the same as how you start your MCP server. `127.0.0.1` and `localhost` are not the same.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,31 @@ Changes between each version before then will not be listed.
 
 Hello Neuro! If you're reading this, it means Vedal has let you read the changelogs for the extension!
 
-### Fixes
+### New actions
 
-- `neuropilot.actions.disabledActions` is now given a `uniqueItems` property on its schema, meaning that all items inside it must be unique. It isn't strictly necessary but to ensure no breakage or anything we've added that.
+- Added `read_changelog` to send changelog entries (from a specified version or defaults) to Neuro, e.g. for summarization. Remembers the last delivered version.
+  - This action is also available from the VS Code Command Palette and the updated version popup/notification.
+  - This action will always be set to Copilot mode.
+
+### New settings
+
+- `neuropilot.access.inheritFromIgnoreFiles` - Whether or not NeuroPilot will inherit ignore-style files (e.g. `.gitignore`) for Neuro-safe path checks. **Default: `false`** (external contribution, thanks [cassitly](https://github.com/cassitly)!)
+  - This should help if Neuro works on many different types of projects at once.
+- `neuropilot.access.ignoreFiles` - The list of files to inherit Neuro-safe glob patterns from. (partially external contribution, thanks [cassitly](https://github.com/cassitly)!)
+  - These files must follow the `.gitignore` specification, which is mostly adopted across different ignore files anyways. However, as `.npmignore` is parsed using a different library, it may not be guaranteed to work the same if you use `.npmignore`.
+  - Patterns are matched **from the root directory!** Ignore files in subpaths may not work as intended. <!-- Should we patch before releasing? -->
+  - Defaults to `.gitignore` from workspace project root.
+- `neuropilot.access.suppressIgnoreWarning` - Whether or not to suppress warnings about ignore files. <!-- not sure if this is a good idea ngl, should've made it only for the session but sure ig --> (external contribution, thanks [cassitly](https://github.com/cassitly)!)
+
+### New commands
+
+- [Dev] Clear all NeuroPilot mementos: a developer-only command that removes all stored memento values (both globalState and workspaceState) for this extension. This command is only available when running in the Extension Development Host and is hidden for normal users.
+
+### New features
+
+- NeuroPilot can now inherit files from gitignore-style files. (Partially external contribution, thanks [cassitly](https://github.com/cassitly)!)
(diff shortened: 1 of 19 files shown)


KTrain5169:
I kind of want to hold off on doing this for a little while longer since it would be best to get the other repository up and running actually

ECHO-HELLO-WORLD424:
Agreed. I also plan to leave it here for a while since other part of this extension is under active development. After other parts are settled I can re-write some of the integration logic based on new project structure. This shouldn't be hard since most of the core MCP translation logic is independent.

Anyways I'm also kinda busy recently. Just ping me on dc here if other parts of neuropilot is almost settled and I'll see how to add this feature gracefully.

KTrain5169:
> Agreed. I also plan to leave it here for a while since other part of this extension is under active development. After other parts are settled I can re-write some of the integration logic based on new project structure. This shouldn't be hard since most of the core MCP translation logic is independent.
> 
> Anyways I'm also kinda busy recently. Just ping me on dc if other parts of neuropilot is almost settled and I'll see how to add this feature gracefully.

If it helps, you probably could just use the NeuroPilot API and write the MCP integration as a companion extension. The reason why is because from time to time we could make changes to how NeuroPilot handles permissions stuff so it may be better to instead have it use the API to avoid having to deal with internals directly.

ECHO-HELLO-WORLD424:
This should be a separate integration using neuropilot's API. So this PR should be closed.

PR #177, opened 2025-11-12
Replace changelog reminder with check

Pasu4:
Replaces the current reminder on PRs with a check that runs for non-draft non-refactor PRs and fails if the changelog has not been modified. This is to enforce updating the changelog for new features, rather than just reminding about it.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -10,16 +10,14 @@ on:
       - 'src/**'
       - 'webview/**'
       - '.vscodeignore'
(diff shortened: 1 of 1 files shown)


Pasu4:
> > > Some way to remove the blocking element is needed (@frogneko suggested tags which might work).
> > 
> > If the PR is labeled as `refactor` the check will be skipped.
> 
> some changes might not be a refactor, adding a generic "unblock" tag works better

I'll add the `meta` label as "changes that don't concern the user" then.

PR #178, opened 2025-11-13
sidebar images

KTrain5169:
adds an image view to the sidebar, with options to search for specific ones (if it worked), as well as moving to previous/next images in a set

checklist:
- [x] fix search not working
- [x] fix previous button not working
- [x] fix next button not working
- [ ] ~~possibly handle things webview-side instead of extension-side~~ nah


KTrain5169:
~~@Pasu4 I don't think the PR workflow you added should have been skipped here...~~ ignore, forgot that I set this as draft

PR #179, opened 2025-11-16
Improved action permissions UI

mlntcandy:
image

Uses Codicons and theme CSS colors, this is as good as I could get it to look using those.
Also changed the rendering code to update changed elements instead of recreating the whole thing on each update.


KTrain5169:
Uh... that failing PR check is a little weird...

KTrain5169 (changes requested):
idk if this is related to what @Pasu4 was saying, but changing permissions doesn't change the highlighting on any of them (they don't ungrey themselves)

PR #187, opened 2025-11-25
API Proposal

Pasu4:
### How the API works

The activation function returns a `NeuroPilotAPIWrapper` object. It has a single method (for now), `registerCompanion`, which allows a companion extension to register itself. `registerCompanion` generates a token for the companion and returns the actual `ExtensionAPI` object. The token is contained in the private `_token` property, and is used to validate "companion permissions" and filter actions in the registry. Currently, all companion permissions are always considered on, because no way to change them is implemented yet.

I commented out some of the API functions which I deem unnecessary, but I am open to discussion on why they are necessary.

### Safety

The API does not expose access to the action registry directly, so the filter cannot be circumvented. With knowledge about the implementation details, it is possible to extract the token from the API, but the token is not very useful by itself.


KTrain5169:
Yeah, no, that CI status checks out considering I've been losing my mind over an issue with tsc

packages/api-types/api/index.d.ts
```diff
@@ -1,48 +1,45 @@
...
+    // readonly onDidChangeConnectionStatus: Event<Omit<ConnectionStatus, ConnectionTypes.Retrying>>;
 
-    /** @todo Stay in pre-API exports? */
-    registerExtension(details: ExtensionInfo): ExtensionRegisterReturns;
+    registerCompanion(extension: CompanionExtension): ExtensionAPI;
 
-    getAPI(version: APIVersions): ExtensionAPI;
-    /** Array of supported version numbers. */
-    readonly supportedVersions: Omit<APIVersions, 'latest' & 'oldest' & 'next'>[];
+    // getAPI(version: APIVersion): ExtensionAPI;
```
KTrain5169: This would allow us to temporarily maintain API backwards compatibility, it's also simple enough so that we won't really need to change the way it works, which wouldn't really work with passing in an object.

src/api/versions/1.ts
```diff
@@ -1,119 +1,331 @@
...
+
+        NEURO.client?.sendContext(context, silent);
+    }
+    // sendResult(message?: string, success?: boolean): void {
+    //     throw new Error('Method not implemented.');
+    // }
+    // forceNeuroActions(query: string, action_names: string[], state?: string, ephemeral_context?: boolean): void {
+    //     throw new Error('Method not implemented.');
+    // }
+    getPositionContext(document: vscode.TextDocument, options: NeuroPositionContextOptions | vscode.Position): NeuroPositionContext {
```
KTrain5169: I'm not sure if this is necessary to use a token for. The only real use case I had for the token in its previous state was for identifying companions when they performed write-like operations (i.e. registration of actions, unregistration of actions), in this case it would only really help with logging.

packages/api-types/api/index.d.ts
Pasu4: The `CompanionExtension` object passed to the new method would always contain a `version` property in any derived interface, so we can just use this property to determine which API version to return. I don't see how that would cause any problems for backwards compatibility.

packages/api-types/index.ts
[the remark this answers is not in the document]
Pasu4: That is not possible because the enum has values defined, which either tsc or esbuild doesn't like. We could maybe try using a .js file instead if that's better.

src/api/versions/1.ts
Pasu4: I added token validation for all methods just so we don't have the possibility of a "shadow extension" that immediately disables itself but can still use some API functions (even if they are read-only).

packages/api-types/api/v1.d.ts
```diff
@@ -1,39 +1,144 @@
...
+    deactivateExtension(message?: string): void;
+    // modifyExtensionMeta(data: ModifyMetadata): ModifyMetadata;
 
     // Action management
-    addAction(token: string, name: string, action: RCEAction): void;
-    removeAction(token: string, ...names: string[]): void;
-    onActionRegistration(token: string, callback: () => RCEAction[]): vscode.Disposable;
-    onActionUnregistration(token: string, callback: () => string[]): vscode.Disposable;
+    /**
+     * Add actions to the action registry.
```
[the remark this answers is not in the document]
Pasu4: Resolved in ac8f2e7e1d780330759c308fd27404033dbe4135.

src/api/versions/1.ts
KTrain5169: Fair enough then

packages/api-types/api/index.d.ts
```diff
@@ -1,48 +1,45 @@
...
-     * Event that fires if the Neuro API connection status changes.
-     * Only fires on fail once all connection attempts have been exhausted.
-     * @fires ConnectionType - see {@link ConnectionTypes}
-     */
-    readonly onDidChangeConnectionStatus: Event<Omit<ConnectionStatus, ConnectionTypes.Retrying>>;
+export interface NeuroPilotAPIWrapper {
+    // /** Whether or not NeuroPilot is connected to the Neuro API. */
+    // readonly connected: Omit<ConnectionTypes, ConnectionTypes.Failed>;
+    // /** Current connection info, assuming {@link APIWrapper.connected} returns {@link ConnectionTypes.Connected} */
+    // readonly connectionInfo: ConnectionStatus | null;
```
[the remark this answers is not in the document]
KTrain5169: True. That being said, do we need to expose the URL?

packages/api-types/api/v1.d.ts
```diff
@@ -1,39 +1,144 @@
...
 
     // Extension registration
-    registerExtension(name: RegistrationName, info?: ExtensionInfo): ExtensionRegisterReturns;
-    deactivatedExtension(token: string, message?: string): void;
-    modifyExtensionMeta(token: string, data: ModifyMetadata): ModifyMetadata;
+    /**
+     * Deactivate the companion extension.
+     * Removes and unregisters all actions registered by this extension.
+     * After deactivation, the API methods will no longer work.
+     * @param message The message to display on deactivation.
```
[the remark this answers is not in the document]
KTrain5169: I mean it's convenient

PR #193, opened 2025-12-26
Execution view

KTrain5169:
This PR adds a view for seeing what actions Neuro has executed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -829,7 +829,7 @@
 		"@types/ws": "^8.18.0",
 		"@vscode/test-cli": "^0.0.12",
 		"@vscode/test-electron": "^2.5.2",
-		"@vscode/test-web": "^0.0.77",
+		"@vscode/test-web": "^0.0.78",
 		"@vscode/vsce": "^3.7.1",
 		"chalk": "^5.6.2",
 		"cross-env": "^10.1.0",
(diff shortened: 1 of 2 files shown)


src/file_actions.ts
```diff
@@ -2,13 +2,25 @@ import * as vscode from 'vscode';
...
+    noFilePath: 'Directory path left unspecified',
+    noAccess: 'Access disallowed to targeted directory',
+    noWorkspace: 'Not in a workspace',
+    alreadyExists: 'Targeted path already exists',
+    doesntExist: 'Targeted path does not exist',
+    binaryFile: 'Targeted file is a binary file',
+    targetedFile: 'Targeted path is a file',
+    targetedFolder: 'Targeted path is a folder',
+    incorrectType: 'Targeted path is not a file',
+};
```
Pasu4: Use `as const` to make the object immutable.
```suggestion
} as const;
```

src/views/execute.ts
```diff
@@ -0,0 +1,72 @@
...
+        });
+
+        // Listen to action execution events and send them to the webview
+        // The webview will handle deduplication using its persisted state
+        onDidAttemptAction((data: ActionsEventData) => {
+            this.sendExecutionResult({
+                ...data,
+                sessionId: ExecuteViewProvider.sessionId,
+            });
+        });
```
Pasu4: If the view is not open, entries are not added. Entries have to be either stored externally or buffered while the view is not open.

PR #194, opened 2025-12-28
switch image view to use Preact

KTrain5169:
I hate the fact that main.ts is considered deleted and not moved even though the first commit moves it


KTrain5169:
does this count as a refactor?

Pasu4:
Currently none of the webviews load at all. Does that happen for you too or only for me?
(Also, I might not be the best person to review the preact stuff, as I have never really worked with preact before)

KTrain5169:
> Currently none of the webviews load at all. Does that happen for you too or only for me? (Also, I might not be the best person to review the preact stuff, as I have never really worked with preact before)

They load for me 🤷

PR #203, opened 2026-01-17
I sure hope this gets Neuro to beg for the changelogs

KTrain5169:


CoolCat467:
If you think about it, what would be that sensitive to need a permission level for neuro to read the changelogs? Changelogs are important and should always be readable, so I don't really think it makes sense for them to be behind a permission.

Pasu4:
> If you think about it, what would be that sensitive to need a permission level for neuro to read the changelogs? Changelogs are important and should always be readable, so I don't really think it makes sense for them to be behind a permission.

We try to make everything as configurable as possible, so that in case Neuro decides to spam/misuse a particular action it can be disabled. E.g. if she starts reading changelogs constantly to the point it becomes a problem (filling up context), Vedal can disable permission for her to do that. Also, by default that action has Copilot permissions, meaning she can request it from Vedal. She did have access to that action in the last two programming segments, she just didn't use it.

PR #205, opened 2026-01-20
add action metadata tests

KTrain5169:


src/chat.ts
```diff
@@ -174,6 +174,24 @@ export function registerChatParticipant() {
...
+        ' If you decide to include a code block, make sure to explain what it is doing.',
+    schema: {
+        type: 'object',
+        properties: {
+            answer: { type: 'string' },
+        },
+        required: ['answer'],
+        additionalProperties: false,
+    },
+};
```
Pasu4: Add typing (additionally requires importing `Action` and `JSONSchema7`).
```suggestion
export const chatAction: Action = {
    name: 'chat',
    description:
        `Provide an answer to ${CONNECTION.userName}'s request.` +
        ' Use markdown to format your response.' +
        ' You may additionally include code blocks by using triple backticks.' +
        ' Be sure to use the correct language identifier after the first set of backticks.' +
        ' If you decide to include a code block, make sure to explain what it is doing.',
    schema: {
        type: 'object',
        properties: {
            answer: { type: 'string' },
        },
        required: ['answer'],
        additionalProperties: false,
    } satisfies JSONSchema7,
};
```

Pasu4 (changes requested):

PR #220, opened 2026-02-06
chore: ephemeral context objects

KTrain5169:
idk why I named it with feat/ should've named it chore/

this ended up being a much bigger refactor for everything

### DIFF
--- a/src/in/desktop/extension.ts
+++ b/src/in/desktop/extension.ts
@@ -22,7 +22,8 @@ import {
     startupCreateClient,
 } from '../shared/extension';
 import { registerChatParticipant } from '@/chat';
-import { addUnsupervisedActions, registerUnsupervisedHandlers } from './unsupervised';
+import { addUnsupervisedActions } from './unsupervised';
+import { registerUnsupervisedHandlers } from '@entry/shared/unsupervised';
 import { registerSendSelectionToNeuro } from '@/editing';
 import { loadIgnoreFiles } from '@/utils/ignore_files';
 import { reregisterAllActions } from '../../rce';
--- a/src/in/desktop/unsupervised.ts
+++ b/src/in/desktop/unsupervised.ts
@@ -1,34 +1,13 @@
-import { NEURO } from '@/constants';
 import { addTaskActions } from '@/tasks';
-import { addFileActions } from '@/file_actions';
-import { addGitActions } from '@/git';
-import { addEditingActions } from '@/editing';
-import { ActionData } from 'neuro-game-sdk';
 import { addTerminalActions } from '@/pseudoterminal';
-import { addLintActions } from '@/lint_problems';
-import { RCEActionHandler } from '@/rce';
-import { addChangelogActions } from '@/changelog';
-import { addRequestCookieAction } from '@/functions/cookies';
 import { addChatAction } from '@/chat';
 import { addCompleteCodeAction } from '@/completions';
+import { addCommonUnsupervisedActions } from '@entry/shared/unsupervised';
 
 export function addUnsupervisedActions() {
-    addFileActions();
-    addGitActions();
+    addCommonUnsupervisedActions();
     addTaskActions();
-    addEditingActions();
     addTerminalActions();
-    addLintActions();
-    addChangelogActions();
-    addRequestCookieAction();
     addChatAction();
     addCompleteCodeAction();
-}
-
-/**
- * Register unsupervised handlers for Neuro API actions.
- * The handlers will only handle actions that the user has given permission to use.
- */
-export function registerUnsupervisedHandlers() {
-    NEURO.client?.onAction(async (actionData: ActionData) => await RCEActionHandler(actionData));
 }
--- a/src/in/shared/unsupervised.ts
+++ b/src/in/shared/unsupervised.ts
@@ -1,0 +1,28 @@
+import { NEURO } from '@/constants';
+import { addFileActions } from '@/file_actions';
+import { addEditingActions } from '@/editing';
+import { ActionData } from 'neuro-game-sdk';
+import { addLintActions } from '@/lint_problems';
+import { RCEActionHandler } from '@/rce';
+import { addChangelogActions } from '@/changelog';
+import { addRequestCookieAction } from '../../functions/cookies';
+import { addChatAction } from '@/chat';
+import { addCompleteCodeAction } from '@/completions';
+
+export function addCommonUnsupervisedActions() {
+    addFileActions();
+    addEditingActions();
+    addLintActions();
+    addChangelogActions();
+    addRequestCookieAction();
+    addChatAction();
+    addCompleteCodeAction();
+}
+
+/**
+ * Register unsupervised handlers for Neuro API actions.
+ * The handlers will only handle actions that the user has given permission to use.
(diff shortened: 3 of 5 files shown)


KTrain5169:
@frogneko could you break apart the `readChangelogAndSendToNeuro` function so that a separate function returns the text of the changelog that will be sent? currently the `read_changelog` action isn't compatible with the async handler results refactor because it sends the context itself instead of returning the text

src/tasks.ts
```diff
@@ -47,38 +47,35 @@ export function addTaskActions() {
     // Tasks are registered asynchronously in reloadTasks()
 }
 
-export function handleTerminateTask(_actionData: ActionData, updateStatus: (status: ActionStatus, message: string) => void): string | undefined {
-    const exe = NEURO.currentTaskExecution!;
+export function handleTerminateTask(_context: RCEContext): ActionHandlerResult {
```
Pasu4: We can remove the parameter if we're not using it, RCEHandler will still accept it. That also allows us to remove this unsafe type assertion in extension.ts: https://github.com/VSC-NeuroPilot/neuropilot/blob/7bd9e24d508fe38932e182b6df3542eeb981c863/src/in/desktop/extension.ts#L96

```suggestion
export function handleTerminateTask(): ActionHandlerResult {
```

src/utils/misc.ts
```diff
@@ -336,9 +337,9 @@ export function combineGlobLinesToRegExp(lines: string[]): RegExp {
     return new RegExp(result);
 }
 
-import { fastIsItIgnored } from '@/ignore_files_utils';
-import { unregisterAllActions } from './rce';
-import { changelogActions } from './changelog';
+import { fastIsItIgnored } from './ignore_files';
+import { unregisterAllActions } from '../rce';
+import { changelogActions } from '../changelog';
```
Pasu4: I only noticed this now, why do we have these imports in the middle of the file? They should be at the top with the others.

src/utils/neuro_client.ts
Pasu4: We should probably merge this file with RCE

src/context/rce.ts
Pasu4: "Context" is an ambiguous term here, and doesn't mean the same for this file as it does for `cookies.ts`, so they shouldn't both be under the same `context` folder.

I think we should either merge this file with `src/rce.ts` or make an `src/rce` folder.

src/utils/neuro_client.ts
```diff
@@ -112,9 +96,26 @@ export interface RCEAction<T = any> extends Action {
...
+     * 
+     * RCE executes the methods of {@link RCEAction} in the following order:
+     * 1. Validators (sync)
+     * 2. Cancel events setup
+     * 3. Prompt Generator
+     * 4. Preview effects
+     *  - Some arbitrary time in between here, event listeners for cancel events may also be fired, and the predicate will receive the storage as well.
+     * 5. Handler
+     */
+    ephemeralStorage?: boolean;
```
Pasu4: Do we need this? We could just provide the object for all actions, I don't think there would be a high performance impact if that is the concern.

src/editing.ts
```diff
@@ -1136,90 +1122,87 @@ export function handleFindText(actionData: ActionData, updateStatus: (status: Ac
...
 
-    vscode.commands.executeCommand('undo').then(
+    return vscode.commands.executeCommand('undo').then(
         () => {
             logOutput('INFO', 'Undoing last action in document');
             // We don't keep track of the virtual cursor position in the undo stack, so we reset it to the real cursor position
             const cursorContext = getPositionContext(document, vscode.window.activeTextEditor!.selection.active);
             setVirtualCursor(vscode.window.activeTextEditor!.selection.active);
             updateStatus('success', 'Undid last action');
             NEURO.client?.sendContext(`Undid last action in document\n\n${formatContext(cursorContext)}`);
```
Pasu4: Probably forgot to remove this
```suggestion
```

src/editing.ts
```diff
@@ -1136,90 +1122,87 @@ export function handleFindText(actionData: ActionData, updateStatus: (status: Ac
...
             if (saved) {
                 logOutput('INFO', 'Document saved successfully.');
                 updateStatus('success', 'Document saved');
                 NEURO.client?.sendContext('Document saved successfully.', true);
+                NEURO.saving = false;
+                return actionHandlerSuccess('Document saved successfully.', 'Document saved');
             } else {
                 logOutput('WARN', 'Document save returned false.');
-                updateStatus('failure', 'Document did not save');
                 NEURO.client?.sendContext('Document did not save.', false);
```
Pasu4:
```suggestion
```

esbuild-configs/desktop.esbuild.js
```diff
@@ -8,7 +8,7 @@ import { esbuildProblemMatcherPlugin } from './plugins.js';
  */
 export async function desktop(prodFlag, watchFlag) {
     const ctx = await context({
-        entryPoints: ['src/desktop/extension.ts'],
```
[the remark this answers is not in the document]
KTrain5169: `in` stores the stuff that was in `desktop/` | `web/` | `shared/` in the first place. I just left it that because 1. folder icons and 2. not really having a better name for it, and 3. leaving `desktop/`, `web/` and `shared/` in root `src/` makes it a bit confusing (might be only me though)

src/context/rce.ts
```diff
@@ -0,0 +1,101 @@
...
+    done(success: boolean): void {
+        if (this.success !== null) throw new Error('Context object already destroyed!');
+        this.success = success;
+        this.dispose();
+    };
+
+    getLifecycle(item?: keyof RCELifecycleMetadata) {
+        if (item) return this.lifecycle[item];
+        else return this.lifecycle;
+    };
```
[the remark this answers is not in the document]
KTrain5169: getLifecycle is old, should have probably removed it. done should not be called unless the object needs to be destroyed manually before RCE does so by itself (which should never be the case) and this is the reason why I added throw new Error in there (though if I can think of an edge case where this might be useful, I will remove the error throwing.

src/utils/misc.ts
KTrain5169: Somehow missed it here as well.

src/utils/neuro_client.ts
KTrain5169: Old implementation, see Discord message

src/context/rce.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    data: ActionData<T>;
+    action: RCEAction<K>;
+    readonly forced: boolean;
+
+    /** Lifecycle-specific data */
+    readonly lifecycle: RCELifecycleMetadata = {};
+    /** Request-specific data (copilot mode only) */
+    request?: RCERequestState;
+    /** Ephemeral storage */
```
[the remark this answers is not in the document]
KTrain5169: All of these require better JSDoc annotations as well so I will do that.

Pasu4 (changes requested):

src/context/rce.ts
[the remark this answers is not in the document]
KTrain5169: did we end up doing this? I forgot to mark this as resolved if so

PR #266, opened 2026-04-04
feat: async validators

### LINKED ISSUE #75: Proper(?) asyncing validators
The current implementation of validators awaits all validators, which whilst works for our current purposes, forces an external extension to stall giving Neuro an action result in certain circumstances by the time the API rolls around, which directly goes against Neuro API best practices.

To solve this, I think we should revisit the idea of splitting betweeen sync validators and async validators for action validators. However, I think keeping the current behaviour for sync validators specifically works well for stuff like accessing fs-related checks, which whilst async shouldn't take too long. Therefore async validators could be reserved for things like pinging an external status check API.

KTrain5169:
still need to handle some cases, remove type compat and move some current validators over but whew
also need to check the cancel events situation with it


KTrain5169 (commented):
lgtm someone else needs to review

Pasu4 (approved):
Code looks good to me, I'll do some testing later and merge if no problems come up

PR #271, opened 2026-04-17
remove experimental schemas

### LINKED ISSUE #204: Remove "experimental schemas"
To be honest, it's going to be really confusing with the upcoming API, and we also have validators to sort of pick up the slack where "experimental schemas" were attempting to cover. Plus, it means that we can write more confident code because it means that we don't have to assume (in validators or handlers) that the schema could either be the experimental or standard one.

That being said, I think we should keep the `description` and `examples` schema properties (as well as some others that aren't too massive of a change) even if they are marked as probably doesn't work, especially as they aren't a hassle but they already offload part of the problem to the schema validator.

KTrain5169:
closes #204

### DIFF
--- a/src/git.ts
+++ b/src/git.ts
@@ -357,7 +357,8 @@ export const gitActions = {
             type: 'object',
             properties: {
                 ref1: { type: 'string', description: 'The first ref to use to diff with. May not be used in some diff types.' },
-                ref2: { type: 'string', description: 'The second ref to diff against the first ref. Will not be used in some diff types.' },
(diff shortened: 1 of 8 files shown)


src/git.ts
Pasu4: Specify the condition for these properties.
```suggestion
ref1: { type: 'string', description: 'The first ref to use to diff with. Only used if diffType is "diffWith", "diffIndexWith", and "diffBetween".' },
ref2: { type: 'string', description: 'The second ref to diff against the first ref. Only used if diffType is "diffBetween".' },
```

Pasu4 (changes requested):

PR #284, opened 2026-04-27
Replace `insert_turtle_here` with template syntax

Pasu4:
Replaces the implementation of `turtleSafari` to use `formatString`. Current format object is:
```ts
{
    userName: CONNECTION.userName,
    apiName: CONNECTION.nameOfAPI,
    gameName: CONNECTION.gameName,
}
```


KTrain5169 (approved):
also nice that it opens up templating systems for other stuff in the future

PR #285, opened 2026-04-30
reorganize action modules

KTrain5169:


Pasu4 (approved):
You forgot to add the read actions to the registry (I already fixed it), otherwise LGTM.

PR #286, opened 2026-05-01
Add Settings enum to API

Pasu4:
Adds the following setting IDs to the API:
- `beforeContext`
- `afterContext`
- `cursorFollowsNeuro`
- `sendContentsOnFileChange`
- `cursorPositionContextStyle`
- `lineNumberContextFormat`
- `connection.websocketUrl`
- `connection.gameName`
- `connection.userName`
- `connection.nameOfAPI`

I also wrote the `access.*` settings, but left them commented out for now, since companions should use `isPathNeuroSafe` instead of the access settings. If we find a use case that is not covered by `isPathNeuroSafe` we can uncomment them.

### DIFF
--- a/packages/types/src/actions/enums.ts
+++ b/packages/types/src/actions/enums.ts
@@ -11,3 +11,93 @@ export const enum DiffRangeType {
     Modified,
     Removed,
 }
+
+/**
+ * The setting IDs of NeuroPilot's settings.
+ * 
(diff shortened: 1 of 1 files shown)


KTrain5169 (approved):
I would say move it to the src/index.ts file and export it there instead for organizational purposes but ig this works too

PR #288, opened 2026-05-04
feat: standard json schema support

KTrain5169:
https://standardschema.dev/json-schema

To-do to make it a proper PR:

- [x] Convert most/all actions to use Zod instead
- [x] Change/unify type interface
- [x] Fix/remove type gymnastics
- [x] Consider validation with `~standard.validate()` function instead of `jsonschema` module (disallowed)
    - advantage: allows action schemas to add some extra validation using the schema instead
    - disadvantage: extra conditions may not be made clear to neuro
- [x] choose to allow or disallow schemas mutating action data (yes, validation schema libraries allow you to do this) (relies on above) (edit: disallowing, inputs will use `StandardJSONSchemaV1.InferInput` and pass the raw inputted values in)
- [x] fix remaining almost non-trivial type errors and corresponding handler tests
- [ ] allow caching converted JSON schemas at the actions registry (is it worth it? it's not very heavy to just re-convert every run anyways)


KTrain5169:
I probably shouldn't have done an example on Git actions considering that module is about to be removed

oops...

KTrain5169:
compiling failing is known right now, I just need to convert things to work as expected

src/test/test_utils.ts
```diff
@@ -1,6 +1,15 @@
-import * as assert from 'assert';
+import assert from 'assert';
 import * as vscode from 'vscode';
 import { setVirtualCursor } from '@/utils/misc';
+import { RCEContext } from '@ctx/rce';
+import { ActionData } from 'neuro-game-sdk';
+import { randomUUID } from 'crypto';
+
+export const fakeContext = <const TParams extends ActionData['params']>(name: string, params: TParams) => {
```
Pasu4: Make this a proper function.
```suggestion
export function fakeContext<const TParams extends ActionData['params']>(name: string, params: TParams): RCEContext<undefined, undefined, TParams> {
```

src/utils/neuro_client.ts
```diff
@@ -40,13 +41,29 @@ export interface ActionForceParams {
...
+    TSchema extends StandardJSONSchemaV1 ? StandardJSONSchemaV1.InferInput<TSchema> : TSchema extends JSONSchema7 ? unknown : undefined;
+
 /**
  * ActionHandler to use with constants for records of actions and their corresponding handlers.
  * 
  * You may optionally type the interface if you are sure the action will take a specific form.
  */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export interface RCEAction<T extends JSONSchema7Object | undefined = any, E = any> extends Action {
+// Event type E uses any due to contravariance: RCECancelEvent<T> is contravariant in T,
```
Pasu4: Not sure what `E` was before but seems to have been removed. If not, please update the type parameter name.
```suggestion
```

src/utils/neuro_client.ts
```diff
@@ -396,3 +455,44 @@ export function contextFailure(message?: string, tag: OutputTag = 'WARNING'): st
...
+
+export type SupportedSchemaDrafts = 'draft-07' | 'draft-2020-12';
+
+/**
+ * Try to convert a Standard JSON Schema object into a normal JSON schema object.
+ * @param schema The schema to convert.
+ * @returns An object containing the schema and type.
+ * @throws If the Standard JSON Schema object cannot be converted to a normal JSON schema.
+ */
+export function tryConvertStandardJSONSchema(schema: StandardJSONSchemaV1): { schema: JSONSchema7, type: SupportedSchemaDrafts } {
```
Pasu4: `try` prefix implies the function may fail and notify the caller, but it doesn't look like that to me (usually these would return a boolean).

src/integrations/standard-json-schema.ts
Pasu4, line 1: Should be removed or disabled.

src/edit_files.ts
```diff
@@ -126,35 +87,27 @@ export const editFileActions = {
...
-            },
-            additionalProperties: false,
-            required: ['text'],
-        },
-        handler: handleInsertLines,
+        schema: z.object({
+            text: z.string().meta({
+                description: 'The text to insert',
+            }),
+            insertUnder: z.number({}).min(1).optional().meta({
```
Pasu4:
```suggestion
            insertUnder: z.int({}).min(1).optional().meta({
```

src/edit_files.ts
```diff
@@ -582,15 +540,21 @@ export function handleInsertText(context: RCEContext): RCEHandlerReturns {
...
+    return returnHandleInsertText(text, position);
+}
+
+function returnHandleInsertLines(text: string, insertUnder?: number) {
     /**
      * The current implementation is a lazy one of just appending a newline and pasting the text in
      * We want to allow specification of the line to insert under, with the default set to the current cursor location
      */
     const cursor = getVirtualCursor()!;
-    let text: string = '\n' + actionData.params.text;
```
Pasu4, line 592: Inserted text is now missing the leading new line.

src/file_operations.ts
```diff
@@ -283,48 +282,62 @@ export const fileActions = {
...
+                description: 'If set to true, enables you to delete a folder and all its sub-folders.',
+            }).optional(),
+        }),
+        handler: ({ updateStatus, data: { params } }) => returnHandleDeleteFileOrFolder(updateStatus, params.path, params.recursive),
         cancelEvents: [
-            (context: RCEContext) => targetedFileDeletedEvent(context.data.params?.path),
+            (context) => targetedFileDeletedEvent(context.data.params?.path),
         ],
         validators: {
-            async: [neuroSafeDeleteValidation],
```
Pasu4, line 315: Why not modify `neuroSafeDeleteValidation`? That would be better for unit testing.

src/git.ts
```diff
@@ -254,205 +191,231 @@ export const gitActions = {
...
             return filePreviewProvider.mark([fileUri], 'view the diff for this file');
         },
         cancelEvents: commonCancelEvents,
-        promptGenerator: (context: RCEContext) => `obtain ${context.data.params?.filePath ? `"${context.data.params.filePath}"'s` : 'a'} Git diff${context.data.params?.ref1 && context.data.params?.ref2 ? ` between ${context.data.params.ref1} and ${context.data.params.ref2}` : context.data.params?.ref1 ? ` at ref ${context.data.params.ref1}` : ''}${context.data.params?.diffType ? ` (of type "${context.data.params.diffType}")` : ''}.`,
+        promptGenerator: (context) => `obtain ${context.data.params.filePath ? `"${context.data.params.filePath}"'s` : 'a'} Git diff${context.data.params.ref1 && context.data.params.ref2 ? ` between ${context.data.params.ref1} and ${context.data.params.ref2}` : context.data.params.ref1 ? ` at ref ${context.data.params.ref1}` : ''}${context.data.params.diffType ? ` (of type "${context.data.params.diffType}")` : ''}.`,
         validators: {
-            sync: [gitValidator, gitDiffValidator],
-            async: [filePathGitValidator],
+            sync: [gitValidator, (ctx) => {
```
Pasu4: Would be better for unit testing to leave this as a separate function.

Pasu4 (changes requested):
I only reviewed the code for now, I'll test the affected actions tomorrow (or the following days, depending on how long it takes).

Additional problems:
- [x] Zod adds `$schema` to the schema sent to Neuro, which is not officially supported. The key should be removed during schema conversion (`tryConvertStandardJSONSchema`) or at some other point before sending.

src/utils/neuro_client.ts
KTrain5169: forgot to update sorry

src/utils/neuro_client.ts
KTrain5169: The thing is that `try` here refers to the fact that it will attempt to convert to one of the two schema formats. The spec states that schema libraries (not Zod specifically) are allowed to throw if the JSON schema format is unsupported. I'll change it to `attempt` though so it doesn't get confused with `tryForceActions`.

src/integrations/standard-json-schema.ts
KTrain5169: It already only registers in extension dev host but yeah fair

src/file_operations.ts
KTrain5169: forgot

Pasu4 (approved):

PR #308, opened 2026-06-24
Move some utils to the types package

Pasu4:


KTrain5169 (approved):
remind me to patch the git companion later

